PR #4, opened 2014-12-17
Add support to flashlight for tcp proxies (fallbacks)

oxtoacart:


src/github.com/getlantern/balancer/dialer.go
```diff
@@ -0,0 +1,135 @@
...
+				log.Trace("Active")
+			}
+			select {
+			case t, ok := <-d.errCh:
+				if !ok {
+					log.Trace("dialer stopped")
+					if d.OnClose != nil {
+						d.OnClose()
+					}
+					return
```
myleshorton: It seems like we either want to make this a break statement or mark the dialer as inactive or both, as I believe the balancer will still consider it active at this point and will keep trying to use it.

oxtoacart: It gets marked inactive here - https://github.com/getlantern/flashlight-build/pull/4/files#diff-66bca3f9f37634c0c70fea788b936ba3R71

oxtoacart: Oh, I misunderstood.  This branch of logic is handling the closing of this dialer, which is signaled by closing errCh, which leads to ok being false.  This is not caused by receiving an error--ok refers to whether or not receiving from the channel was ok.

myleshorton: Right. In that case, though, the dialer won't be marked as inactive and will still be used as far as I can tell.

myleshorton: Although I suppose that only happens when the balancer itself is closed, which should presumably mean it's fine.

oxtoacart: Yeah, but this is only called by the Close() method on the balancer itself, after which point the balancer shouldn't be used anymore.

Just to be super safe, I updated the Close() method on the balancer to no longer reference closed dialers.

https://github.com/getlantern/flashlight-build/commit/2b001ca252457bbbf9b2bc2f8a62943fc4d6f331

oxtoacart: And for completeness, I just updated the unit test - https://github.com/getlantern/flashlight-build/commit/2b37e3da42aefdce6cefdb7cbce159b79091ff64

src/github.com/getlantern/bytecounting/bytecounting_conn.go
```diff
@@ -0,0 +1,58 @@
+// package bytecounting provides mechanisms for counting bytes read/written on
+// net.Conn and net.Listener.
```
myleshorton: nit: isn't it really just for counting bytes on a net.Conn, but you can use it from a net.Listener to count bytes on an accepted net.Conn?

oxtoacart: No, the package also provides an implementation of net.Listener that wraps the connections from a standard net.Listener with the byte counting functionality.

https://github.com/getlantern/flashlight-build/blob/tcpproxying/src/github.com/getlantern/bytecounting/bytecounting_listener.go

src/github.com/getlantern/chained/server.go
```diff
@@ -0,0 +1,65 @@
...
+
+	connIn, _, err := hj.Hijack()
+	if err != nil {
+		log.Errorf("Unable to hijack connection: %s", err)
+		connOut.Close()
+		return
+	}
+
+	go io.Copy(connOut, connIn)
+	go io.Copy(connIn, connOut)
```
myleshorton: What happens when either the client or the destination closes the connection? I would think the other side should always be closed as well.

myleshorton: So I would think we'd want to check for Copy returning and then close the corresponding connection.

oxtoacart: Great catch!  Fixed here - https://github.com/getlantern/flashlight-build/commit/e6e011d663e0dce3b111f9eb658efcdd459fef27

src/github.com/getlantern/flashlight/client/balancer.go
```diff
@@ -0,0 +1,52 @@
...
+	} else {
+		log.Trace("Creating balancer channel")
+		client.balCh = make(chan *balancer.Balancer, 1)
+	}
+	log.Trace("Publishing balancer")
+	client.balCh <- bal
+	client.balInitialized = true
+
+	return bal
+}
```
myleshorton: I know in practice the caller for this wraps everything in mutex, but it seems like it might make more sense to do that here for all of initBalancer where the mutex is really needed.

src/github.com/getlantern/flashlight/client/chained.go
```diff
@@ -0,0 +1,92 @@
...
+		sessionCache := tls.NewLRUClientSessionCache(1000)
+		dial = func() (net.Conn, error) {
+			conn, err := tlsdialer.DialWithDialer(netd, "tcp", s.Addr, false, &tls.Config{
+				ClientSessionCache: sessionCache,
+				InsecureSkipVerify: true,
+			})
+			if err != nil {
+				return nil, err
+			}
+			if !conn.ConnectionState().PeerCertificates[0].Equal(x509cert) {
```
myleshorton: So it looks like the first certificate in the chain here is the last in the chain. Might it not make more sense to match against the root signing certificate?

src/github.com/getlantern/flashlight/client/chained.go
```diff
@@ -0,0 +1,92 @@
...
+				conn.Close()
+				return nil, fmt.Errorf("Server's certificate didn't match expected!")
+			}
+			return conn, err
+		}
+	}
+
+	d := &chained.Dialer{
+		DialServer: dial,
```
myleshorton: I'd be tempted to make this a little more self-documenting and to create a certPinningDialer function that this calls. Then this would just be:

```go
d := &chained.Dialer{
		DialServer: certPinningDialer(),
```

src/github.com/getlantern/flashlight/client/chained.go
oxtoacart: This just replicates how it works in Lantern today.  We only ever use self-signed certificates.

src/github.com/getlantern/flashlight/client/balancer.go
oxtoacart: The mutex is needed at a larger scope, so why not just keep it in Client.Configure()?

src/github.com/getlantern/flashlight/client/chained.go
oxtoacart: Agreed, will do it.

src/github.com/getlantern/flashlight/server/server.go
```diff
@@ -131,73 +102,40 @@ func (server *Server) Configure(newCfg *ServerConfig) {
...
+		WriteTimeout:               server.WriteTimeout,
+		CertContext:                server.CertContext,
+		AllowNonGlobalDestinations: server.AllowNonGlobalDestinations,
 	}
 
-	if server.host != "" {
-		log.Debugf("Running as host %s", server.host)
+	if server.cfg.Unencrypted {
+		log.Debug("Running in unencrypted mode")
+		fs.CertContext = nil
```
myleshorton: When would we want to allow this? Testing?

src/github.com/getlantern/flashlight/genconfig/cloud.yaml.tmpl
```diff
@@ -1,7 +1,7 @@
 # cloud.yaml contains the default configuration that's made available on the
 # internet.
 client:
-  servers:
+  frontedservers:
```
myleshorton: So this will break compatibility with old clients right? Is the plan to just version the file name again?

src/github.com/getlantern/flashlight/client/chained.go
oxtoacart: So, after playing with this, I realized that moving it into its own function doesn't buy us much.  We still need a factory-type function to load the certificate and initialize the session cache, and by moving that stuff into its own function we just add some boilerplate and additional error handling clauses.

oxtoacart: ^ That is to say, I think it's best to keep it as it is.

src/github.com/getlantern/flashlight/server/server.go
oxtoacart: Correct.

src/github.com/getlantern/flashlight/genconfig/cloud.yaml.tmpl
oxtoacart: Correct.  See comment on https://github.com/getlantern/lantern/pull/2103.

src/github.com/getlantern/flashlight/client/chained.go
oxtoacart: So which cert in the chain we verify depends on which cert was distributed to us via kscope.  To fully future proof it, we would probably want to walk the whole chain and accept the peer if any of the certs matches the specified.  But since we don't really know the exact scenario and requirements, I say we leave it for now.

src/github.com/getlantern/balancer/.travis.yml
```diff
@@ -0,0 +1,14 @@
+language: go
+
+go:
+  - 1.3.3
```
myleshorton: Do they support 1.4 yet?

oxtoacart: Oh yeah, most of our travis configs were already on 1.4, I just missed a few somehow.

Fixed - https://github.com/getlantern/flashlight-build/commit/a362d2cf49a8907b70217440c3f1a37ba885b4a8

myleshorton:
So this generally looks good to me @oxtoacart. I'm wondering if we want to do another flashlight build prior to merging this to fix the 100% CPU bug on the client. What do you think?

oxtoacart:
@myleshorton Thanks for reviewing!  Yes, I think it makes sense to put out a release with the CPU usage fix.  I anticipated that we might want to do this as a hot fix, so that PR actually went to master.  We can do a new build from there anytime, irrespective of when you merge tcpproxying to devel.

src/github.com/getlantern/flashlight/server/server.go
oxtoacart: To elaborate - we support unencrypted running on the server so that we can sit behind @aranhoide 's TLS terminator.  I made the client able to connect directly to such a server for testing purposes.

myleshorton:
I'm not seeing the PR to master. In Lantern?

oxtoacart:
> I'm not seeing the PR to master. In Lantern?

You already merged it some time ago - https://github.com/getlantern/flashlight-build/pull/19

teamlantern:
Oh nice forgot about that. OK cool so we should probably put out a hotfix.
It's tempting to include the configureNetworkServices change in there too.

On Fri, Jan 9, 2015 at 9:36 AM, oxtoacart <notifications@github.com> wrote:

> I'm not seeing the PR to master. In Lantern?
>
> You already merged it some time ago - #19
> <https://github.com/getlantern/flashlight-build/pull/19>
>
> —
> <https://github.com/getlantern/flashlight-build/pull/4#issuecomment-69368893>
> .
>



-- 
--
President
https://www.getlantern.org

oxtoacart:
Yes, I think the configureNetworkServices change would be a good one to put in there too.  For the hotfix, we will need to produce a flashlight build to bundle with Lantern.

oxtoacart:
@myleshorton Is it okay to go ahead and merge this now?

myleshorton:
Merging!

oxtoacart:
Thanks!

PR #18, opened 2015-01-05
getlantern/lantern#2135 Switched to using net.SplitHostAndPort

oxtoacart:


src/github.com/getlantern/flashlight/server/server.go
```diff
@@ -234,15 +235,12 @@ func determineInternalIP() (string, error) {
...
+	return host, err
 }
 
 func onBytesGiven(destAddr string, req *http.Request, bytes int64) {
-	port := "0"
-	parts := strings.Split(destAddr, ":")
-	if len(parts) > 1 {
-		port = parts[1]
-	}
+	_, port, _ := net.SplitHostPort(destAddr)
```
myleshorton: So this is slightly different now in the case of a quirky address, as port will be the empty string instead of 0. Is that what you want?

oxtoacart: Good catch!  No, it's not what I want, I still want 0.  Will fix.

src/github.com/getlantern/flashlight/server/server.go
```diff
@@ -234,14 +235,14 @@ func determineInternalIP() (string, error) {
...
 }
 
 func onBytesGiven(destAddr string, req *http.Request, bytes int64) {
-	port := "0"
-	parts := strings.Split(destAddr, ":")
-	if len(parts) > 1 {
-		port = parts[1]
+	_, port, _ := net.SplitHostPort(destAddr)
+	if port == "" {
+		port = "0"
```
oxtoacart: @myleshorton Ensures port is 0 instead of blank.

myleshorton: Nice.

myleshorton:
OK LGTM! Merging.

oxtoacart:
@myleshorton Thanks!

PR #19, opened 2015-01-05
2121

oxtoacart:


myleshorton:
LGTM

myleshorton:
You want me to merge or do you have more plans for this branch?

oxtoacart:
@myleshorton Thanks for taking a look.  It turns out that this didn't fix the problem, so I'm still toying around with it.  I'll assign it to you once I'm comfortable that it actually fixes the problem.

src/github.com/getlantern/enproxy/lazyconn.go
```diff
@@ -46,6 +46,8 @@ func (l *lazyConn) get() (conn net.Conn, err error) {
 
 		// Wrap the connection in an idle timing one
 		l.connOut = idletiming.Conn(conn, l.p.IdleTimeout, func() {
+			l.p.connMapMutex.Lock()
+			defer l.p.connMapMutex.Unlock()
```
oxtoacart: @myleshorton This was a drive-by.  I noticed during testing that we weren't synchronizing access to this map on deleting a conn.

src/github.com/getlantern/idletiming/idletiming_conn.go
```diff
@@ -167,12 +176,12 @@ func (c *IdleTimingConn) SetDeadline(t time.Time) error {
 }
 
 func (c *IdleTimingConn) SetReadDeadline(t time.Time) error {
-	c.readDeadline = &t
+	atomic.StoreInt64(&c.readDeadline, t.UnixNano())
```
oxtoacart: @myleshorton Switched to using atomic.StoreInt64 to avoid concurrency issues on writing/reading the deadline.

src/github.com/getlantern/idletiming/idletiming_conn.go
```diff
@@ -82,14 +85,16 @@ func (c *IdleTimingConn) TimesOutAt() time.Time {
 // Read implements the method from io.Reader
 func (c *IdleTimingConn) Read(b []byte) (int, error) {
 	totalN := 0
+	readDeadline := time.Unix(0, atomic.LoadInt64(&c.readDeadline))
```
oxtoacart: @myleshorton I now grab the deadline up-front before entering the loop.  This way, if the client happens to change the deadline while I'm in the middle of looping, it won't have unpredictable effects.

src/github.com/getlantern/idletiming/idletiming_conn.go
```diff
@@ -100,12 +105,13 @@ func (c *IdleTimingConn) Read(b []byte) (int, error) {
 			n, err := c.conn.Read(b)
 			c.markActive(n)
 			totalN = totalN + n
-			timedOut := isTimeout(err)
-			if timedOut {
-				// Ignore timeouts when using deadline based on IdleTimeout
+			hitMaxDeadline := isTimeout(err) && !time.Now().Before(maxDeadline)
```
oxtoacart: !time.Now().Before(maxDeadline) is important.  This controls for the situation where the read/write times out not because of the deadline that we set above but for some other reason.  In that case, we want to just go ahead and return the timeout error and break out of the loop.

myleshorton: Isn't the only timeout our own timeout (i.e. don't the connections not have a timeout)? What would another reason be? I guess we don't actually know what type of net.Conn it may be (might even be an idletiming_conn)?

oxtoacart: Yeah, ultimately we don't have control over the underlying connection and what's done with it.  Really, all these changes are just me trying to nitpick the heck out of this code to see if I can get it to a point where it doesn't break in production.  I never was able to reproduce in unit testing.

Note - fl-singapore-004-1 has been running spike free for about 23 minutes now, which is not bad relative to previous tests.  If it still looks good in the morning, I'll push the updated code to all fallbacks and see what happens.

myleshorton: OK gotcha. Yeah these changes look fine to me. Merging!

oxtoacart:
@myleshorton Thanks for reviewing!

PR #20, opened 2015-01-06
Fixes to idletiming.Conn to make sure it closes after idling

oxtoacart:
This will prevent leaking file descriptors on the flashlight server.

### DIFF
--- a/src/github.com/getlantern/enproxy/lazyconn.go
+++ b/src/github.com/getlantern/enproxy/lazyconn.go
@@ -49,6 +49,7 @@ func (l *lazyConn) get() (conn net.Conn, err error) {
 			l.p.connMapMutex.Lock()
 			defer l.p.connMapMutex.Unlock()
 			delete(l.p.connMap, l.id)
+			conn.Close()
 		})
 	}
 
--- a/src/github.com/getlantern/idletiming/idletiming_conn.go
+++ b/src/github.com/getlantern/idletiming/idletiming_conn.go
@@ -16,13 +16,14 @@ var (
 // the specified period. Read and Write calls will timeout if they take longer
 // than the indicated idleTimeout.
 //
-// Once a connection is determined to be idle, it is closed.
-//
 // idleTimeout specifies how long to wait for inactivity before considering
 // connection idle.
 //
 // onClose is an optional function to call after the connection has been closed,
 // whether or not that was due to the connection idling.
+//
+// Note - idletiming.Conn does not close the underlying connection, it is up to
+// clients to handle this in their onClose callback.
 func Conn(conn net.Conn, idleTimeout time.Duration, onClose func()) *IdleTimingConn {
 	c := &IdleTimingConn{
 		conn:             conn,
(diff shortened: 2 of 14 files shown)


oxtoacart:
@myleshorton I know I've already assigned a lot of PRs to you, sorry about that.  They're all in the area of flashlight server stability, which you've been interested in, thus I'm assigning you as reviewer.  If you want to get some help on any of the PRs, please feel free to reassign.

src/github.com/getlantern/enproxy/lazyconn.go
```diff
@@ -49,6 +49,7 @@ func (l *lazyConn) get() (conn net.Conn, err error) {
 			l.p.connMapMutex.Lock()
 			defer l.p.connMapMutex.Unlock()
 			delete(l.p.connMap, l.id)
+			conn.Close()
 		})
```
myleshorton: Where did this used to get closed?

oxtoacart: Exactly :)

src/github.com/getlantern/idletiming/idletiming_conn.go
```diff
@@ -157,7 +159,10 @@ func (c *IdleTimingConn) Write(b []byte) (int, error) {
 }
 
 func (c *IdleTimingConn) Close() error {
-	c.closedCh <- true
+	select {
+	case c.closedCh <- true:
+		// close accepted
+	}
```
myleshorton: Why the select here?

oxtoacart: To avoid blocking if Close is called multiple times.

oxtoacart: Oh wait, I need a default clause!  Hang on.

src/github.com/getlantern/idletiming/idletiming_listener.go
```diff
@@ -15,20 +15,29 @@ import (
...
 	idleTimeout time.Duration
-	onClose     func()
+	onClose     func(conn net.Conn)
 }
 
 func (l *idleTimingListener) Accept() (c net.Conn, err error) {
 	c, err = l.orig.Accept()
 	if err == nil {
-		c = Conn(c, l.idleTimeout, l.onClose)
+		var closeFn func()
```
myleshorton: Seems like this should be var closeFn = func(){} as otherwise you'll get a nil reference if l.onClose == nil and cloneFn is never assigned.

src/github.com/getlantern/idletiming/idletiming_conn.go
```diff
@@ -156,8 +159,15 @@ func (c *IdleTimingConn) Write(b []byte) (int, error) {
...
 }
 
+// Close this IdleTimingConn. This will close the underlying net.Conn as well,
+// however any error from closing that connection is NOT returned here.
 func (c *IdleTimingConn) Close() error {
-	c.closedCh <- true
+	select {
+	case c.closedCh <- true:
+		// close accepted
+	default:
```
myleshorton: Nice.

src/github.com/getlantern/idletiming/idletiming_listener.go
oxtoacart: That's what I want, as Conn only calls closeFn() if it isn't nil.

myleshorton: OK. Seems like it would be a little less error prone to just have it be a function that does nothing though, especially since I think if we ever did screw up and call it when it's nil it would result in a panic right?

myleshorton:
LGTM!

PR #22, opened 2015-01-07
Listen/openssl (POC, don't merge)

aranhoide:
**Don't merge this!**; this PR was only made to make it easier to see what was changed in this proof-of-concept implementation of flashlight listening via OpenSSL.


aranhoide:
I'll kill this PR but leave the branch around, since it might come in handy if and when we ever try and do Chrome mimicry.

PR #23, opened 2015-01-08
Track new stats distinctClients and distinctDestHosts

oxtoacart:
For getlantern/lantern#2148

Note - this required statshub changes which were made in https://github.com/getlantern/statshub/commit/ab494cc76b2b34c5a4925da3948b2d57576bb382

### DIFF
--- a/src/github.com/getlantern/flashlight/server/server.go
+++ b/src/github.com/getlantern/flashlight/server/server.go
@@ -241,7 +241,7 @@ func determineInternalIP() (string, error) {
 }
 
 func onBytesGiven(destAddr string, req *http.Request, bytes int64) {
-	addr, port, _ := net.SplitHostPort(destAddr)
+	host, port, _ := net.SplitHostPort(destAddr)
 	if port == "" {
 		port = "0"
 	}
@@ -257,7 +257,7 @@ func onBytesGiven(destAddr string, req *http.Request, bytes int64) {
 		givenTo := statreporter.Country(clientCountry)
 		givenTo.Increment("bytesGivenTo").Add(bytes)
 		givenTo.Increment("bytesGivenToByFlashlight").Add(bytes)
-		givenTo.Member("distinctDestAddrs", addr)
+		givenTo.Member("distinctDestHosts", host)
 
 		clientIp := req.Header.Get("X-Forwarded-For")
 		if clientIp != "" {
@@ -268,5 +268,4 @@ func onBytesGiven(destAddr string, req *http.Request, bytes int64) {
 		}
 
 	}
-
 }
--- a/src/github.com/getlantern/flashlight/server/server.go
+++ b/src/github.com/getlantern/flashlight/server/server.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"reflect"
 	"strconv"
+	"strings"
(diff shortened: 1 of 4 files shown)


src/github.com/getlantern/flashlight/server/server.go
```diff
@@ -240,7 +241,7 @@ func determineInternalIP() (string, error) {
 }
 
 func onBytesGiven(destAddr string, req *http.Request, bytes int64) {
-	_, port, _ := net.SplitHostPort(destAddr)
+	host, port, _ := net.SplitHostPort(destAddr)
```
atavism: I notice we don't check for a non-nil error here, but we do in other places where we call SplitHostPort. I haven't traced the call graph, but is that because destAddr is guaranteed to be of a valid format when we arrive here?

src/github.com/getlantern/flashlight/statreporter/statreporter.go
```diff
@@ -147,11 +147,23 @@ func (r *reporter) run() {
 					categoryStats = make(stats)
 					dgAccum.categories[update.category] = categoryStats
 				}
-				switch update.action {
+				switch a := update.action.(type) {
```
atavism: I had to do some googling to figure out exactly what was going on here, which led me to http://blog.denevell.org/golang-interface-type-assertions-switch.html -- I didn't even know about type switches! This is super cool and useful when you're working with an interface{} type.

src/github.com/getlantern/flashlight/statreporter/statreporter.go
```diff
@@ -196,8 +208,22 @@ func (dgAccum *dimGroupAccumulator) makeReport() report {
...
-		report[category] = accum
+	for category, s := range dgAccum.categories {
+		if category == members {
+			// Transform maps into arrays
+			s2 := make(stats)
+
+			for k, v := range s {
+				m := v.(map[string]bool)
+				a := make([]string, 0, len(m))
+				for member, _ := range m {
```
atavism: You could actually drop the blank identifier here since the value doesn't have to be accounted for.. 
for member := range m { ... }

src/github.com/getlantern/flashlight/server/server.go
oxtoacart: We actually don't care whether this succeeds or not.  If it fails, we get a blank host and 0 port, both of which we record anyway.

src/github.com/getlantern/flashlight/statreporter/statreporter.go
oxtoacart: Yeah, it's pretty handy sometimes.  It's not as cool as the full on pattern matching that's available in some languages, but it's useful.

src/github.com/getlantern/flashlight/statreporter/statreporter.go
oxtoacart: Yes it is.  I've removed the unnecessary assignment, thanks!

oxtoacart:
@atavism I've addressed your comments.  Look good otherwise?

atavism:
Yep! Merging.

oxtoacart:
@atavism Thanks!

PR #25, opened 2015-01-12
getlantern/lantern#2162 Restricted outbound access to just ports 80, 443...

oxtoacart:
..., 8080 and 8443

### DIFF
--- a/src/github.com/getlantern/fronted/server.go
+++ b/src/github.com/getlantern/fronted/server.go
@@ -6,6 +6,7 @@ import (
 	"net"
 	"net/http"
 	"os"
+	"strconv"
 	"time"
 
 	"github.com/getlantern/enproxy"
@@ -64,6 +65,10 @@ type Server struct {
 	// AllowNonGlobalDestinations: if false, requests to LAN, Loopback, etc.
 	// will be disallowed.
 	AllowNonGlobalDestinations bool
+
+	// AllowedPorts: if specified, only connections to ports listed in this
+	// slice will be allowed.
+	AllowedPorts []int
 
 	// OnBytesSent: optional callback for learning about bytes sent by this
 	// server to upstream destinations.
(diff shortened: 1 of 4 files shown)


src/github.com/getlantern/fronted/server.go
```diff
@@ -169,27 +174,63 @@ func (server *Server) Serve(l net.Listener) error {
...
 }
 
+func (server *Server) checkForLocalAddress(addr string) error {
+	host, _, err := net.SplitHostPort(addr)
+	if err != nil {
+		return fmt.Errorf("Unable to split host and port for %v: %v", addr, err)
+	}
+
+	ipAddr, err := net.ResolveIPAddr("ip", host)
+	if err != nil || len(ipAddr.IP) == 0 {
```
oxtoacart: This was a drive-by fix.  I noticed that if passing in something like ":0", ResolveIPAddr would not return an error, but line 206 would fail because IP is empty.

myleshorton: Cool. Maybe a better name would be checkForGlobalUnicast?

src/github.com/getlantern/fronted/server.go
```diff
@@ -169,27 +174,63 @@ func (server *Server) Serve(l net.Listener) error {
...
+
+func (server *Server) checkForDisallowedPort(addr string) error {
+	_, port, err := net.SplitHostPort(addr)
+	if err != nil {
+		return fmt.Errorf("Unable to split host and port for %v: %v", addr, err)
+	}
+
+	portAllowed := false
+	for _, allowed := range server.AllowedPorts {
+		if port == strconv.Itoa(allowed) {
```
myleshorton: Maybe slightly better to do this outside of the loop, as in -- not a biggie of course, but since every request hits this:

```go
port := strconv.Atoi(portString)
for _, allowed := range server.AllowedPorts {
		if port == allowed {
```

src/github.com/getlantern/fronted/server.go
oxtoacart: Renamed to checkForNonGlobalDestination.

src/github.com/getlantern/fronted/server.go
oxtoacart: Fixed.

myleshorton:
LGTM.

PR #26, opened 2015-01-12
Android NDK bindings

xiam:
# Summary

This pull request provides code changes that allow cross compiling a flashlight binary for Android hosts and also a shared library (`libgojni.so`) for using flashlight with the Android's NDK.
## How to build the Android library

Get the `golang/mobile` [docker image](https://github.com/golang/mobile):

```
docker pull golang/mobile
```

Install gobind if you don't already have it:

```
go install golang.org/x/mobile/cmd/gobind
```

Clone my `flashlight-build` repo and switch to the `experimental/android-ndk-bindings` branch.

```
git clone git@github.com:xiam/flashlight-build.git
```

Change into the new `flashlight-build` directory, check out the experimental branch and set `GOPATH` to `$PWD`.

```
cd flashlight-build
git checkout -b experimental/android-ndk-bindings remotes/origin/experimental/android-ndk-bindings
export GOPATH=$PWD
```

Get the `x/mobile` package (see the [documentation](https://github.com/golang/mobile)).

```
go get -d golang.org/x/mobile/example/...
```

Install gobind if you don't already
Generate the Go and Java bindings using `gobind`:

```
cd $GOPATH/src/github.com/getlantern/flashlight/android
./bind.bash
```

Go back to the `GOPATH` library and use the docker image to build the `.so` library:

```
cd $GOPATH
docker run -v $PWD/src:/src golang/mobile /bin/bash -c \ "cd /src/github.com/getlantern/flashlight/android/ && ./make.bash"

find | grep '\.so$'
# ./src/github.com/getlantern/flashlight/android/libs/armeabi-v7a/libgojni.so
```
## How to create a test Android project

Open Android Studio and choose "Import Non-Android studio project"

screenshot from 2015-01-12 15 55 21

Browse to the `.../flashlight/android` directory after building the `libgojni.so` library and click OK.

screenshot from 2015-01-15 17 45 08

Set an import directory and click "Next".

screenshot from 2015-01-12 15 56 46

Wait for Android Studio to finish importing and restart when asked.

Add a new MainActivity by right-clicking on the android module and selecting New->Activity->Blank Activity.

screenshot from 2015-01-12 16 06 19

Add two buttons (with id `stopProxyButton` and `startProxyButton`). Associate the startProxyButton's onclick event to the `startProxyButtonOnClick()` public function and the stopProxyButton's onclick event to the `stopProxyButtonOnClick` function.

screenshot from 2015-01-15 17 41 11

Overwrite the `MainActivity.java` code with the following code:

``` java
package org.getlantern.example;

import go.Go;
import go.flashlight.Flashlight;
import android.app.Activity;
import android.os.Bundle;
import android.util.Log;
import android.view.View;
import android.widget.Button;
import org.getlantern.example.R;


public class MainActivity extends Activity {

    private Thread proxyThread = null;
    private Runnable proxyRunnable = null;
    private Button killButton;
    private Button startButton;

    @Override
    protected void onCreate(Bundle savedInstanceState) {
        super.onCreate(savedInstanceState);

        setContentView(R.layout.activity_main);

        // Initializing application context.
        Go.init(getApplicationContext());

        killButton = (Button)findViewById(R.id.stopProxyButton);
        startButton = (Button)findViewById(R.id.startProxyButton);

        // Disabling stop button.
        killButton.setEnabled(false);

        // Enabling proxy button.
        startButton.setEnabled(true);
    }

    public void stopProxyButtonOnClick(View v) {
        if (proxyThread.isAlive()) {
            Log.v("DEBUG", "Attempt to stop running proxy.");
            Flashlight.StopClientProxy();


            // Disabling stop button.
            killButton.setEnabled(false);

            // Enabling proxy button.
            startButton.setEnabled(true);

            // Asking the thread to interrupt itself.
            proxyThread.interrupt();
            proxyThread = null;
        }
    }

    public void startProxyButtonOnClick(View v) {
        Log.v("DEBUG", "Attempt to run client proxy on :9192");

        if (proxyThread == null || proxyThread.isInterrupted()) {
            proxyRunnable = new Runnable() {
                @Override
                public void run() {
                    try {
                        Log.v("DEBUG", "Running proxy...");
                        Flashlight.RunClientProxy("0.0.0.0:9192");
                    } catch (Exception e) {
                        throw new RuntimeException(e);
                    }
                }
            };

            // Creating a new thread so the UI does not block.
            proxyThread = new Thread(proxyRunnable);

            proxyThread.start();

        } else {
            Log.v("DEBUG", "Proxy is already running");
        }

        // Enabling stop button.
        killButton.setEnabled(true);

        // Disabling proxy button.
        startButton.setEnabled(false);
    }
}
```

Deploy to a real device, when the app starts touch the Start button to initialize the proxy, see the output in the "Android" tab of the logs panel.

screenshot from 2015-01-12 16 32 37

As long as the app is open, you'll be able to test the canonical example by finding the device's IP and proxy through it:

```
curl -x 10.10.100.97:9192 http://www.google.com/humans.txt 
# Google is built by a large team of engineers, designers, researchers, robots, and others in many different sites across the globe. It is updated continuously, and built with more tools and technologies than we can shake a stick at. If you'd like to help us out, see google.com/careers.
```

If you don't want other devices to proxy throw the phone, modify the parameter given to `RunClientProxy()`.


src/github.com/getlantern/flashlight/.gitignore
```diff
@@ -7,3 +7,6 @@ flashlight.ex?
 *.pyc
 certs/errors.txt
 flashlight.yaml
```
xiam: Use the `bind.bash` script to generate those files.

src/github.com/getlantern/flashlight/android/AndroidManifest.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="utf-8"?>
```
xiam: We are including this file following instructions from https://github.com/golang/mobile so the user can bootstrap a test application easily.

src/github.com/getlantern/flashlight/android/bind.bash
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
```
xiam: This file creates Java and Go bindings using gobind. See https://godoc.org/golang.org/x/mobile/cmd/gobind

src/github.com/getlantern/flashlight/android/bindings/main.go
```diff
@@ -0,0 +1,56 @@
+// package flashlight provides minimal configuration for spawning a flashlight
```
xiam: This package defines the exported methods that are going to be available to Android developers using the NDK.

src/github.com/getlantern/flashlight/android/bindings/main_test.go
```diff
@@ -0,0 +1,16 @@
+package flashlight
```
xiam: A very basic test file that can be used to test the `android` build tag (go test -tags="android") and the exported methods without having to deploy the `.so` everytime.

src/github.com/getlantern/flashlight/android/build.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
```
xiam: Build instructions following documentation from https://github.com/golang/mobile

src/github.com/getlantern/flashlight/android/main.go
```diff
@@ -0,0 +1,13 @@
+// main package holds references to the go_bindings package and is based on
```
xiam: This file links our actual bindings with the Go's mobile package and is the basis for building the `.so` library. See https://godoc.org/golang.org/x/mobile/cmd/gobind

src/github.com/getlantern/flashlight/android/make.bash
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
```
xiam: Script that is going to be followed by the docker image.

src/github.com/getlantern/flashlight/flashlight.go
```diff
@@ -136,38 +134,6 @@ func runClientProxy(cfg *config.Config) {
 	}
 }
```
xiam: This function was moved to `server.go` and `server_stubs.go` so it can be disabled on android hosts based on build tags.

src/github.com/getlantern/flashlight/server/server.go
```diff
@@ -138,93 +135,6 @@ func (server *Server) ListenAndServe() error {
 	return fs.Serve(l)
 }
```
xiam: This function was moved to `nattywad.go` and `nattywad_stubs.go` so it can be disabled on android hosts based on build tags.

src/github.com/getlantern/go-natty/natty/natty.go
```diff
@@ -37,7 +37,11 @@ var (
 func init() {
 	nattyBytes, err := bin.Asset("natty")
 	if err != nil {
```
xiam: We're logging instead of panicking to allow Android hosts (where natty is disabled by build tags) to continue working.

myleshorton:
This is looking great @xiam. I made a couple of edits to the instructions here and there, and I'm wondering if our versions of Android Studio might differ slightly. I'm using 1.0.2, and it looks like you were using 1.0.1. In particular, when you added an activity, did you do it through right clicking on the android module and then New->Activity->Blank Activity as I changed in the comments? That seems to generally work OK, although Android Studio didn't automatically generate my R.java file, so I had to fiddle a little bit on that front (Android newbie over here). I also unfortunately don't have a live phone to test on, but I'll fix that shortly.

Otherwise, I think it would be fantastic if that whole Summary were added as a separate markdown file in the flashlight project that we could just link to from the main README. It also seems like there could be some consolidation of the setup into a larger setup.bash script, something like that. It could do things like check for docker and just things like 

```
which gobind || go install golang.org/x/mobile/cmd/gobind"
```

It seems like it should even be possible to script a good deal of the whole setup. In general, though, great stuff!

xiam:
@myleshorton,

You're right about the New->Activity->Blank activity instruction to generate the new activity but I didn't have a problem with the `R.java` file, not sure if this is a problem on my side, I'm also new to Android Studio. If you're unable to complete the build I can check again to make sure I didn't miss something important.

What do you think about adding a `README` file on the `android/` subdirectory? I think I can do that and I can also create a `Makefile` to automate the generation of the `libgojni.so` file. I'll make sure this `Makefile` works on both Linux and OSX.

oxtoacart:
@xiam @myleshorton 

Sorry about only thinking of this now, but what would you think about moving the android support into its own project and just importing the necessary dependencies from flashlight?  I think that might be a little cleaner than an android subdirectory.

To be clear, you can still keep this in flashlight-build, but rather than having it in src/github.com/getlantern/flashlight/android, it would be something like src/github.com/getlantern/flashlight-android.

Part of my thinking here is that as we move more of the desktop app functionality into flashlight, it and the android version will probably diverge significantly.

oxtoacart:
> what would you think about moving the android support into its own project

I should also mention that this is not urgent.  Since we're trying to hit Friday for having the POC wrapped up, I would suggest deferring anything like this until after that.

xiam:
@oxtoacart The android package can be moved around with no major problems, we'd only need to update the import path but, I'm a bit worried about the build tags on the rest of flashlight-build, some packages (such as nattywad or server) had to be changed (or splitted) in order for certain features to be disabled on the android build, so these changes would still be required even if we move the android subdirectory. Do you think we need this fixed for the friday delivery? If that's the case I could create another fresh branch of the current `getlantern:tcpproxying` and integrate my changes trying to leave `flashlight` as clean as possible and creating a new `github.com/getlantern/flashlight-android` repo to put the tools for building `libgojni.so`, it should not take that long as I already know what to do. If this is not strictly necessary now, we could discuss a better strategy after the PoC is delivered.

oxtoacart:
@xiam 

> The android package can be moved around with no major problems, we'd only need to update the import path but, I'm a bit worried about the build tags on the rest of flashlight-build, some packages (such as nattywad or server) had to be changed (or splitted) in order for certain features to be disabled on the android build, so these changes would still be required even if we move the android subdirectory.

That's totally fine.

> Do you think we need this fixed for the friday delivery?

No, it can wait.

> If that's the case I could create another fresh branch of the current getlantern:tcpproxying and integrate my changes trying to leave flashlight as clean as possible and creating a new github.com/getlantern/flashlight-android repo to put the tools for building libgojni.so, it should not take that long as I already know what to do. If this is not strictly necessary now, we could discuss a better strategy after the PoC is delivered.

Let's focus on the PoC first.  We can figure this stuff out later.

myleshorton:
> Let's focus on the PoC first. We can figure this stuff out later.

Agreed. For future reference, I think we should ultimately create a "lantern-android" project that probably just calls all the sub-projects directly and skips flashlight itself entirely. That should allow it to live completely independently with the sub-projects continuing to fulfill their more precise roles applicable to both top-level projects.

myleshorton:
Oh, and I think a Makefile sounds great @xiam!

xiam:
> Let's focus on the PoC first. We can figure this stuff out later.

Agreed. Let me know if you need anything else for this delivery.

> Oh, and I think a Makefile sounds great @xiam!

OK! I'll start working on a build script (already outside `flashlight`) on a separate branch and we'll get back to this discussion later.

myleshorton:
> Agreed. Let me know if you need anything else for this delivery.

I think the only other thing that would be nice is this documentation living on its own *somewhere* that's not the pull request itself and probably that's not flashlight. Actually, I just went ahead and created a lantern-android repository that for now can be just the readme and can point to your fork if that's OK. We'll figure out the larger organizational elements later.

myleshorton:
Oh the new repository is at:

https://github.com/getlantern/lantern-android

xiam:
@myleshorton I'm currently having some trouble on the middle of the build script we were talking about, this problem is related to the import paths lantern-android is expecting. I'm sure I can fix this problem easily if I had a branch on `getlantern/flashlight-build` were I can move things around (since it's the import path lantern-android would expect). Would it be OK if I just created the branch with a name like `experimental/lantern-android` on this repository to keep on working?

Also, I'm a bit unsure on how are we supposed to update upstream repositories, such as https://github.com/getlantern/flashlight from changes we are doing here on `flashlight-build`, should I have any special considerations to ensure minimal friction for upstream?

xiam:
Actually I don't need to "move things around", I think it sounds more dangerous than it actually is, I just need to merge my pull request (and delete the android subdirectory) to keep on working.

xiam:
Please see the docs (wip, still untested on OSX) in https://github.com/getlantern/lantern-android/tree/feature/documentation-and-build-script, I'm supposing I'll be able to create a `experimental/lantern-android` branch.

When the Android build flags are integrated into mainstream we could get rid of the `$GOPATH` hack and leave everything to the `Makefile` script.

myleshorton:
Hey sorry for the delay @xiam. It looks like you worked everything out, or do you still need to create that branch? That would be totally fine, and I'll make sure you have permissions in the relevant repos (I think you should already actually, but I'll double check).

myleshorton:
Yeah you should have the appropriate permissions @xiam!

xiam:
@myleshorton Thanks! I ended up creating the branch to continue working and I just pushed it after your confirmation.

I updated the https://github.com/getlantern/lantern-android with instructions and a `Makefile` script to build the shared library for Android with minimum effort (except the part of manually cloning repositories, we'll have to do it that way until my changes to the build tags hit upstream).

I also created a branch `experimental/lantern-android` and removed the `android` subdirectory from `flashlight-build`.

I think I should add pictures for the Android Studio part, just like I did here in the pull request, but maybe I should invest time on some other task. What do you think?

myleshorton:
Thanks @xiam! Oh right you probably copied those pics in. I actually think they're extremely helpful and that it would be worth adding them to the README if you have time. You could use the same ones too with just saving them from the web page if you don't already have them saved.

xiam:
@myleshorton I actually didn't want to upload the pictures to the repository, so I'm hotlinking them from here :-).

I think it looks better now.

myleshorton:
Oh nice cool -- yeah looks awesome! I'll send it over to the greatfire guys -- great stuff @xiam. There will likely be some followup with this in the coming weeks, but for now we might have you look at more pieces of the port to go. Oh, also, it's a national holiday here on Monday, so we won't have our normal call. If you want to touch base directly, though, just let me know.

xiam:
@myleshorton Sure, let's talk later :). Enjoy your holiday! In the meantime, I'll review the issues list and choose another one, any suggestions?

xiam:
Good news!

This merge is not required anymore, I complied [libflashlight](https://github.com/getlantern/lantern-android) against the `tcpproxying` branch of [flashlight-build](https://github.com/getlantern/flashlight-build) and achieved the same functionality by replicating specific parts of the flashlight client.

Few things are pending though, see the [corresponding issue](https://github.com/getlantern/lantern-android/issues/1).

PR #34, opened 2015-01-21
Pull refactored peerscanner and enproxy

oxtoacart:
Lots of changes.

1. enproxy is significantly cleaned up

2. peerscanner has been refactored to run as a single process on a regular vps (e.g. Digital Ocean)

peerscanner can be deployed using Salt per the changes in https://github.com/getlantern/lantern_aws/pull/34

This related lantern_aws pull request adds support for deploying peerscanner to Digital Ocean and updates flashlight fallbacks to register with peerscanner - https://github.com/getlantern/lantern_aws/pull/34


.travis.yml
```diff
@@ -9,6 +9,8 @@ install:
   - go get -v github.com/mattn/goveralls
 
 script:
+  # This unencryptes envvars.bash.enc into envvars.bash so that we can run the tests
+  - openssl aes-256-cbc -K $encrypted_263604b5b445_key -iv $encrypted_263604b5b445_iv -in envvars.bash.enc -out envvars.bash -d
```
oxtoacart: Decrypts environment variables for unit test that connects to CloudFlare.

src/github.com/getlantern/balancer/.travis.yml
```diff
@@ -1,7 +1,7 @@
 language: go
 
 go:
-  - 1.4
+  - 1.4.1
```
oxtoacart: I went ahead and upgraded all the builds to Go 1.4.1

src/github.com/getlantern/balancer/balancer_test.go
```diff
@@ -110,7 +110,7 @@ func TestAll(t *testing.T) {
 		QOS:    15,
 		Dial: func(network, addr string) (net.Conn, error) {
 			atomic.StoreInt32(&dialedBy, 3)
-			if checkAttempts < 6 {
+			if atomic.LoadInt32(&checkAttempts) < 6 {
```
oxtoacart: Cleaned up some race conditions that caused this test to occasionally fail.

src/github.com/getlantern/cloudflare/.gitignore
```diff
@@ -0,0 +1,23 @@
+# Compiled Object files, Static and Dynamic libs (Shared Objects)
```
oxtoacart: I brought the cloudflare project into flashlight-build.

src/github.com/getlantern/cloudflare/api.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+
+	return req, nil
+
+}
+
+// decodeBody is used to JSON decode a body
+func decodeBody(resp *http.Response, out interface{}) error {
+	body, err := ioutil.ReadAll(resp.Body)
+	resp.Body.Close()
```
oxtoacart: This is a change that's intended to help with the file descriptor leak.

src/github.com/getlantern/cloudflare/record.go
```diff
@@ -0,0 +1,351 @@
...
+		params["ttl"] = "1"
+	}
+
+	req, err := c.NewRequest(params, "POST", "rec_new")
+	if err != nil {
+		return nil, err
+	}
+
+	resp, err := checkResp(c.Http.Do(req))
+	if resp != nil && resp.Body != nil {
```
oxtoacart: This and similar changes are intended to help with the file descriptor leak.

src/github.com/getlantern/enproxy/conn_impl.go
```diff
@@ -17,8 +18,8 @@ import (
 // addr: the host:port of the destination server that we're trying to reach
 //
 // config: configuration for this Conn
-func Dial(addr string, config *Config) (*Conn, error) {
-	c := &Conn{
+func Dial(addr string, config *Config) (net.Conn, error) {
```
oxtoacart: Dial now returns the net.Conn interface rather than a concrete *Conn.

src/github.com/getlantern/enproxy/conn_impl.go
```diff
@@ -17,8 +18,8 @@ import (
...
+	c := &conn{
```
oxtoacart: Conn is now conn (no longer exported).

src/github.com/getlantern/enproxy/conn_impl.go
```diff
@@ -38,12 +37,18 @@ func Dial(addr string, config *Config) (*Conn, error) {
 		return nil, fmt.Errorf("Unable to dial proxy: %s", err)
 	}
 
+	go c.processWrites()
+	go c.processReads()
 	go c.processRequests(proxyConn)
```
oxtoacart: All goroutines are now spawned after the initial dial to the proxy.  That way, if the initial dial failed and we return an error, we don't have to clean up after them.

src/github.com/getlantern/enproxy/conn_impl.go
```diff
@@ -38,12 +37,18 @@ func Dial(addr string, config *Config) (*Conn, error) {
...
 
-	return c, nil
+	return idletiming.Conn(c, c.config.IdleTimeout, func() {
+		c.Close()
+		// Close the initial proxyConn just in case
+		proxyConn.conn.Close()
+	}), nil
```
oxtoacart: enproxy.conn used to do its own idle timing.  To simplify the code, I removed that and wrap it with an idletiming.Conn instead.  idletiming is well unit tested, so it's safer to use than the old enproxy logic which wasn't as thoroughly unit tested.  Also, it's nice to separate these concerns.

src/github.com/getlantern/enproxy/conn_impl.go
```diff
@@ -153,19 +166,6 @@ func (c *Conn) doRequest(proxyConn *connInfo, host string, op string, request *r
...
-	c.lastActivityTime = time.Now()
-}
-
-func (c *Conn) isIdle() bool {
-	c.lastActivityMutex.RLock()
-	defer c.lastActivityMutex.RUnlock()
-	timeSinceLastActivity := time.Now().Sub(c.lastActivityTime)
-	return timeSinceLastActivity > c.config.IdleTimeout
-}
-
```
oxtoacart: Used by old idling logic, no longer needed.

src/github.com/getlantern/enproxy/conn_impl_reads.go
```diff
@@ -4,101 +4,86 @@ import (
...
-	err := initialResponse.err
-	if err != nil {
-		return
-	}
-
-	proxyHost := initialResponse.proxyHost
-	proxyConn := initialResponse.proxyConn
-	resp := initialResponse.resp
+func (c *conn) processReads() {
+	increment(&reading)
```
oxtoacart: You'll see a lot of these calls to increment and decrement.  This is used to generate the periodic metrics printout for debugging connections.

src/github.com/getlantern/enproxy/conn_impl_reads.go
```diff
@@ -107,39 +92,14 @@ func (c *Conn) processReads() {
 
 // submitRead submits a read to the processReads goroutine, returning true if
 // the read was accepted or false if reads are no longer being accepted
-func (c *Conn) submitRead(b []byte) bool {
-	c.readMutex.RLock()
-	defer c.readMutex.RUnlock()
-	if c.doneReading {
+func (c *conn) submitRead(b []byte) bool {
+	c.closingMutex.RLock()
```
oxtoacart: We used to lock on the closed status.  It turns out that this could cause deadlocks.  In practice, we don't need to know if closing has completed, we just need to know that we're closing.

src/github.com/getlantern/enproxy/conn_intf.go
```diff
@@ -200,60 +205,105 @@ func (c *Conn) Write(b []byte) (n int, err error) {
...
-// Close() implements the function from net.Conn
-func (c *Conn) Close() error {
-	c.closedMutex.Lock()
-	defer c.closedMutex.Unlock()
-	if !c.closed {
-		c.stopReadCh <- nil
-		c.stopWriteCh <- nil
-		c.stopRequestCh <- nil
-		c.closed = true
+func (c *conn) fail(err error) {
```
oxtoacart: We used to have some code paths that would fail and leave the connection hanging.  This new fail method allows them to report the error in a way that readers and writers can find out about it, and the connection is closed.

src/github.com/getlantern/enproxy/conn_test.go
```diff
@@ -62,18 +60,80 @@ func TestBadBuffered(t *testing.T) {
 	doTestBad(true, t)
 }
 
+func TestIdle(t *testing.T) {
```
oxtoacart: Makes sure that idling closes open connections.

src/github.com/getlantern/fdcount/fdcount.go
```diff
@@ -41,16 +40,23 @@ func (c *Counter) AssertDelta(expected int) error {
 	if err != nil {
 		return err
 	}
-	actual := c.countMatches(out) - c.startingCount
+	endingLines, endingCount := c.matchingLines(out)
```
oxtoacart: fdcount is a utility for counting file descriptors.  I made some changes to make the messages that it reports more useful.

src/github.com/getlantern/flashlight/client/fronted.go
```diff
@@ -53,6 +53,7 @@ func (s *FrontedServerInfo) dialer(masqueradeSets map[string][]*fronted.Masquera
 	fd := fronted.NewDialer(&fronted.Config{
 		Host:               s.Host,
 		Port:               s.Port,
+		PoolSize:           30,
```
oxtoacart: Fronted now requires pooling to be explicitly configured.

src/github.com/getlantern/flashlight/config/flags.go
```diff
@@ -20,6 +20,7 @@ var (
 	statsPeriod    = flag.Int("statsperiod", 0, "time in seconds to wait between reporting stats. If not specified, stats are not reported. If specified, statshub, instanceid and statsaddr must also be specified.")
 	statshubAddr   = flag.String("statshub", "pure-journey-3547.herokuapp.com", "address of statshub server")
 	instanceid     = flag.String("instanceid", "", "instanceId under which to report stats to statshub. If not specified, no stats are reported.")
+	registerat     = flag.String("registerat", "", "base URL for peer DNS registry at which to register (e.g. https://peerscanner.getiantem.org)")
```
oxtoacart: Support for registering with peerscanner directly.

src/github.com/getlantern/flashlight/flashlight.go
```diff
@@ -61,16 +57,8 @@ func main() {
...
-		startCPUProfiling(cfg.CpuProfile)
-		defer stopCPUProfiling(cfg.CpuProfile)
-	}
-
-	if cfg.MemProfile != "" {
-		defer saveMemProfile(cfg.MemProfile)
-	}
-
-	saveProfilingOnSigINT(cfg)
+	finishProfiling := profiling.Start(cfg.CpuProfile, cfg.MemProfile)
```
oxtoacart: I moved profiling into a new utility library that I can use in multiple programs.

src/github.com/getlantern/flashlight/server/server.go
```diff
@@ -137,9 +142,43 @@ func (server *Server) ListenAndServe() error {
 	if err != nil {
 		return fmt.Errorf("Unable to listen at %s: %s", server.Addr, err)
 	}
+
+	go server.register()
+
 	return fs.Serve(l)
 }
 
+func (server *Server) register() {
```
oxtoacart: Implementation of registration with peerscanner.

src/github.com/getlantern/flashlight/server/server.go
```diff
@@ -258,5 +297,15 @@ func onBytesGiven(destAddr string, req *http.Request, bytes int64) {
...
 		givenTo.Increment("bytesGivenTo").Add(bytes)
 		givenTo.Increment("bytesGivenToByFlashlight").Add(bytes)
+		givenTo.Member("distinctDestHosts", host)
+
+		clientIp := req.Header.Get("X-Forwarded-For")
+		if clientIp != "" {
+			// clientIp may contain multiple ips, use the first
+			ips := strings.Split(clientIp, ",")
+			clientIp := strings.TrimSpace(ips[0])
+			givenTo.Member("distinctClients", clientIp)
```
oxtoacart: New stats for distinct hosts and clients.

src/github.com/getlantern/keyman/keyman.go
```diff
@@ -293,3 +299,48 @@ func bytesToCert(derBytes []byte) (*Certificate, error) {
...
 }
+
+/*******************************************************************************
+ * Utility Functions
+ ******************************************************************************/
+
+// StoredPKAndCert returns a PK and certificate for the given host, storing
+// these at the given pkfile and certfile paths and using the stored values on
+// subsequence calls.
+func StoredPKAndCert(pkfile string, certfile string, organization string, name string) (*PrivateKey, *Certificate, error) {
```
oxtoacart: New function for servers like peerscanner that need certs.

src/github.com/getlantern/peerscanner/cf/cf.go
```diff
@@ -0,0 +1,101 @@
+// package cf provides a utility for interacting with CloudFlare
+package cf
```
oxtoacart: This logic used to be in common.

src/github.com/getlantern/peerscanner/cf/cf.go
```diff
@@ -0,0 +1,101 @@
...
+)
+
+var (
+	log = golog.LoggerFor("cf")
+)
+
+type Util struct {
+	Client *cloudflare.Client
+	domain string
+}
```
oxtoacart: Note that we don't bother caching records anymore.  We don't need to since the application maintains state that includes the records.

src/github.com/getlantern/peerscanner/cf/cf_test.go
```diff
@@ -0,0 +1,47 @@
+package cf
```
oxtoacart: New unit test for CloudFlare utility.

src/github.com/getlantern/peerscanner/host.go
```diff
@@ -0,0 +1,357 @@
...
+
+	proxiedClient *http.Client
+}
+
+func (h *host) String() string {
+	return fmt.Sprintf("%v (%v)", h.name, h.ip)
+}
+
+// newHost creates a new host for the given name, ip and optional DNS record.
+func newHost(name string, ip string, record *cloudflare.Record) *host {
```
oxtoacart: This is the run loop for each host actor.  This is the most critical logic to understand.

src/github.com/getlantern/peerscanner/host.go
```diff
@@ -0,0 +1,357 @@
...
+
+	return h
+}
+
+// run starts the main run loop for this host on a goroutine
+func (h *host) run() {
+	go h.doRun()
+}
+
+func (h *host) doRun() {
```
oxtoacart: This is the run loop for each host actor. This is the most critical logic to understand.

src/github.com/motain/gocheck/.bzr/README
```diff
@@ -0,0 +1,3 @@
+This is a Bazaar control directory.
```
oxtoacart: Dependency of CloudFlare library.

src/github.com/getlantern/profiling/profiling.go
```diff
@@ -0,0 +1,105 @@
+// package profiling provides support for easily doing CPU and memory profiling
+// from within Go programs.
+package profiling
```
oxtoacart: New utility library to help with profiling.

oxtoacart:
This now seems to be fully working on prod and is ready for review.

oxtoacart:
@fffw This is the pull request that we were talking about.

src/github.com/getlantern/connpool/connpool.go
```diff
@@ -46,11 +49,12 @@ type Pool struct {
 	runMutex   sync.Mutex
 	running    bool
 	actualSize int
```
fffw: `actualSize` is assigned as `p.Size` when `Start()` but never changes afterwards.

In fact, no matter how large wg is in `Stop()`, [L153](https://github.com/getlantern/flashlight-build/blob/peerscanner/src/github.com/getlantern/connpool/connpool.go#L153) will `wg.Done()` it.

src/github.com/getlantern/connpool/connpool.go
```diff
@@ -20,17 +20,20 @@ var (
...
+// expire after ClaimTimeout. Lazily here means that the Pool won't start to
+// fill until the first request for a connection. As long as the Pool is
+// actively being used, it will attempt to always have Size number of
+// connections ready to use.
 type Pool struct {
-	// MinSize: the pool will always attempt to maintain at least these many
+	// Size: while active, the pool will attempt to maintain at these many
 	// connections.
-	MinSize int
+	Size int
```
fffw: As changing `Size` after `Start()` has no impact on how connpool works, we'd better disallow changing at all.

Pass it as a parameter to `Start()`, and rename `actualSize` to `size`, or initialise it through `NewPool()`.

src/github.com/getlantern/connpool/connpool.go
```diff
@@ -20,17 +20,20 @@ var (
...
 
 	// ClaimTimeout: connections will be removed from pool if unclaimed for
-	// longer than ClaimTimeout.  The default ClaimTimeout is 10 minutes.
+	// longer than ClaimTimeout. The default ClaimTimeout is 10 minutes. Once
+	// connections have been removed from the Pool, they won't be replaced until
+	// another connection is requested, at which point the pool will fill again.
 	ClaimTimeout time.Duration
```
fffw: Although changing `ClaimTimeout` does affect runtime behaviour, it is not supposed to be changed, and it's not atomic. Same for `RedialDelayIncrement` and `MaxRedialDelay`.

src/github.com/getlantern/balancer/balancer_test.go
```diff
@@ -132,7 +132,7 @@ func TestAll(t *testing.T) {
 		Dial: func(network, addr string) (net.Conn, error) {
 			atomic.StoreInt32(&dialedBy, 4)
 			defer atomic.AddInt32(&d4attempts, 1)
-			if d4attempts < 1 {
+			if atomic.LoadInt32(&d4attempts) < 1 {
 				// Fail once
 				return nil, fmt.Errorf("Me no dialee")
```
xiam: Not a big deal, but I think this error string is a bit confusing, this is an intentional error, so something like "Failing intentionally." would be more appropriate.

oxtoacart: Fixed

src/github.com/getlantern/cloudflare/api.go
```diff
@@ -0,0 +1,117 @@
...
+// request was successful. A non-200 request returns an error
+// formatted to included any validation problems or otherwise
+func checkResp(resp *http.Response, err error) (*http.Response, error) {
+	// If the err is already there, there was an error higher
+	// up the chain, so just return that
+	if err != nil {
+		return resp, err
+	}
+
+	switch i := resp.StatusCode; {
```
xiam: I think this switch statement is a bit confusing: it includes an instruction (`i := resp.StatusCode`) and it may be not clear at first read that it actually has no control variable (nothing after the `;`), so all subsequent cases must generate a boolean value. I think this would be clearer:

```go
i := resp.StatusCode

switch {
case i == 200:
  ...
}
```

And we could replace the `200` constant for `http.StatusOK`.

Also, I'm not totally sure why this switch statement must evaluate against a boolean value (maybe a plan to include more conditionals), nor of the utility of copying `resp.StatusCode` into `i`, but at this time, this would be equivalent to:

```
switch res.StatusCode {
case http.StatusOK:
   return resp, nil
}

return resp, fmt.Errorf("API Error: %s", resp.Status)
```

oxtoacart: @xiam The cloudflare library isn't actually one of ours, it's just something we forked to make a few fixes.  I haven't touched most of it.

xiam: Ooh, got it @oxtoacart :-)! I'll do a quick code review of it anyways, just in case, but I'll consider it as an external resource.

src/github.com/getlantern/cloudflare/api.go
xiam: Awesome finding! I'll also look for things like this.

src/github.com/getlantern/cloudflare/record.go
```diff
@@ -0,0 +1,351 @@
...
+	req, err := c.NewRequest(*params, "GET", "rec_load_all")
+
+	if err != nil {
+		return nil, fmt.Errorf("Error creating request: %s", err)
+	}
+
+	resp, err := checkResp(c.Http.Do(req))
+	if resp != nil && resp.Body != nil {
+		defer func() {
+			resp.Body.Close()
```
xiam: Using defer like this could lead to race conditions, actually not in this case, unless we set `resp.Body` to `nil` before returning. @oxtoacart I think I could fix small and easy things like this after commenting the code, it totally won't be a problem, should I?

src/github.com/getlantern/cloudflare/api.go
oxtoacart: @xiam I think that's a good idea.  If you find anything that's actually broken, we should probably fix it.

src/github.com/getlantern/cloudflare/record.go
oxtoacart: > Using defer like this could lead to race conditions, actually not in this case, unless we set resp.Body to nil before returning.

Yeah, this should be safe.  However it would be less verbose if I had just said `defer resp.Body.Close()`.

> I think I could fix small and easy things like this after commenting the code, it totally won't be a problem, should I?

Yes, although please also point them out as part of the review so that I can learn from them!

src/github.com/getlantern/fdcount/fdcount.go
```diff
@@ -41,16 +40,23 @@ func (c *Counter) AssertDelta(expected int) error {
...
+		return fmt.Errorf("Unexpected TCP file descriptor count. Expected %d, have %d.\n\n%s",
+			expected, actual, lsofDelta(c.startingLines, endingLines))
 	}
 	return nil
 }
 
-func (c *Counter) countMatches(out []byte) int {
-	return bytes.Count(out, []byte(c.match))
+func (c *Counter) matchingLines(out []byte) (string, int) {
+	lines := make([]string, 0)
```
xiam: @oxtoacart  Is there an average or a maximum number of lines we could expect to get from this function? Do you think it would have any practical benefit if we extend the capacity of the `lines` array to prevent reallocations by `append()`?

oxtoacart: This is a testing utility, so it's not performance critical.  Not worth worrying about IMHO.

src/github.com/getlantern/peerscanner/cf/cf.go
```diff
@@ -0,0 +1,97 @@
...
+	return &Util{client, domain}
+}
+
+func (util *Util) GetRotationRecords(subdomain string) ([]cloudflare.Record, error) {
+	recs, err := util.GetAllRecords()
+	if err != nil {
+		return nil, fmt.Errorf("Could not get records:", err)
+	}
+
+	rotation := make([]cloudflare.Record, 0)
```
xiam: We could probably add a capacity to `rotation` too, if that represented a performance gain or a clear benefit (if we expected lots of appends or to call this function frequently). Otherwise I agree we could just skip the note.

oxtoacart: Good point.  When I went to look at this method, I realized that it's actually no longer used.  So I removed it.  100% performance improvement :)

src/github.com/getlantern/peerscanner/host.go
```diff
@@ -0,0 +1,407 @@
...
+
+	// Now actually try to proxy an http request
+	site := testSites[rand.Intn(len(testSites))]
+	resp, err := h.proxiedClient.Head("http://" + site)
+	if err != nil {
+		return false, false, fmt.Errorf("Unable to make proxied HEAD request to %v: %v", site, err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != 200 && resp.StatusCode != 301 {
```
xiam: Not a top priority or something to be fixed right now but we could use constants such as  `http.StatusOK` or `http.StatusMovedPermanently` to both improve readability and prevent typos, not only for cases like these but also for string constants, for example writing `req.Header.Get(headerXForwardedFor)` instead of `req.Header.Get("X-Forwarded-For")` to be absolutely sure constant values are in fact constants. I mean, if we had an error writing the constant `headerXForwardedFor` the compiler would tell us immediately, but if we had a typo while writing "X-Forwarded-For" it would be not as easy to find, as the compiler can't tell of typos on values.

Of course, this is not something we should worry about right now, maybe a task I can take in the future.

oxtoacart: > Not a top priority or something to be fixed right now but we could use constants such as http.StatusOK or http.StatusMovedPermanently

Yes!  Believe it or not, I didn't even realize the http package defined those constants until recently.

> I mean, if we had an error writing the constant headerXForwardedFor the compiler would tell us immediately, but if we had a typo while writing "X-Forwarded-For" it would be not as easy to find, as the compiler can't tell of typos on values.

I'm a little less convinced of this.  In this particular project, we're only referring to that string in one place.  If we made it a constant, we can still have a typo in the constant.  And instead of seeing the typo right where the string is used, you have to go jump to a constant to find it.  My rule of thumb on this is that if it's used more than once, make it a constant, otherwise don't.

xiam: >  My rule of thumb on this is that if it's used more than once, make it a constant, otherwise don't.

Absolutely! maybe we could look for good candidates to be promoted to constants later.

xiam:
I did a "quick" dive into the diff and found no real issues, but I located some interesting pieces that I'd like to review and test in deep: conpool, enproxy and peerscanner, of course. What's our deadline for this PR?

oxtoacart:
@xiam 

> I did a "quick" dive into the diff and found no real issues, but I located some interesting pieces that I'd like to review and test in deep: conpool, enproxy and peerscanner, of course. What's our deadline for this PR?

I'd like to get this into devel before we cut the 1.6.0 release.  I'll let @myleshorton chime in on timeline and priorities relative to that.

If we're in a hurry, one approach would be to:

1. Do some more testing of this branch
2. Merge it
3. Continue the code review while we start testing the 1.6.0 release

myleshorton:
Hey sorry for the delay in responding guys, but I think a thorough review is really more important than pushing 1.6.0 out the door. Actually, with @oxtoacart and I on vacation at the end of this week, it could be a really nice opportunity to really take your time with the libraries you mentioned @xiam. I think it would be super valuable to have you up to speed on those libraries.

So let's just push 1.6.0 until next week, and take your time with the review.

oxtoacart:
> I think a thorough review is really more important than pushing 1.6.0 out the door

+1

src/github.com/getlantern/connpool/connpool_test.go
```diff
@@ -105,10 +106,7 @@ func TestDialFailure(t *testing.T) {
 			}
 			return net.DialTimeout("tcp", addr, 15*time.Millisecond)
 		},
-	}
-
-	p.Start()
-	defer p.Stop()
+	})
```
fffw: connpool will not dial until we Get() a new conn
```go
	// Pool starts to fill
	_, err = p.Get()
	if err == nil {
		t.Fatal("p.Get() should fail here as Dial() func fails")
	}
```

fffw:
Currently, `connpool` only redials after p.Get() invoked, which seems a bit late.

Say, when network unavailable temporarily, all goroutines will fail to dial and wait for `freshenCh`, caller will receive error and may wait until next event. But connpool will not try redial during this period, until next p.Get() invoked.

I made a change [here](https://github.com/getlantern/flashlight-build/commit/c05efffeb965cba7b01b32cb82bfeb56f2978503) under branch peerscanner-fffw, using a timer to auto-redial during waiting for `freshenCh`.

`export TRACE=TRUE; go test` will reveal the effect of this change.

oxtoacart:
@fffw

Thanks for looking into connpool and for the suggestions!

> connpool will not dial until we Get() a new conn

Yes, that's intentional.  It's actually something that I only changed recently in response to #2185.  We used to dial eagerly, and I suspect that it was causing CloudFlare to think that we were DDOS'ing it, since during idle times we would open connections, do a TLS handshake and then eventually close them without ever reading or writing anything.  That's one way of doing a DDOS attack (although it's even more efficient if the client doesn't complete the TLS handshake).

Get() will never wait for connpool to fill, it just dials on its own, so there's no danger of the client blocking while it waits for a redial - https://github.com/getlantern/flashlight-build/blob/peerscanner/src/github.com/getlantern/connpool/connpool.go#L115

> Currently, connpool only redials after p.Get() invoked, which seems a bit late.

Looking it again, I realize that there's little point in redialing periodically unless we know that we've had a successful dial again.  For example, if network connectivity is lost, all of the dials done in the feedConn loop will just continue to fail.  So, to reduce the complexity of the code and not waste resources trying to dial when the network is down, I removed the redial functionality.  Once we get into a state where connections are failing, every call to Get() will dial its own connection.  As soon as one of those dials succeeds, we call freshen which reactivates all of the feedConn loops.

> I made a change here under branch peerscanner-fffw, using a timer to auto-redial during waiting for freshenCh.

I like that this change took out the time.Sleep, since that used to block closing unnecessarily.  By taking out the redial logic, I was able to remove the Sleep call as well.

oxtoacart:
@fffw I noticed that you used timer.Stop() - https://github.com/getlantern/flashlight-build/blob/peerscanner-fffw/src/github.com/getlantern/connpool/connpool.go#L156

Does that prevent a resource leak?  I haven't been doing that.  Should I?

fffw:
>  I noticed that you used timer.Stop() - https://github.com/getlantern/flashlight-build/blob/peerscanner-fffw/src/github.com/getlantern/connpool/connpool.go#L156

>  Does that prevent a resource leak? I haven't been doing that. Should I?

I found its usage at https://github.com/getlantern/idletiming/blob/master/idletiming_conn.go#L41, and think it might be a good practice, as a running timer may still holds OS resource.

Here's a [relevant discussion thread](https://groups.google.com/forum/#!topic/golang-nuts/Chx1tCs2QGg).

oxtoacart:
@fffw Thanks!  So it looks like a timer will consumer some resources until its next timeout is hit.  I'll comb through my use of timers and see if any need a Close().

xiam:
> Actually, with @oxtoacart and I on vacation at the end of this week, it could be a really nice opportunity to really take your time.

Hey @myleshorton, @oxtoacart have fun! I'll continue discussing the code with @fffw, in case I need anything really important I'll reach you by mail.

src/github.com/getlantern/connpool/connpool.go
```diff
@@ -9,9 +9,7 @@ import (
 )
 
 const (
-	DefaultClaimTimeout         = 10 * time.Minute
-	DefaultRedialDelayIncrement = 50 * time.Millisecond
-	DefaultMaxRedialDelay       = 1 * time.Second
+	DefaultClaimTimeout = 10 * time.Minute
```
xiam: I looked for these exported constants `DefaultClaimTimeout`, `DefaultRedialDelayIncrement`, `DefaultMaxRedialDelay` and didn't find any piece of code that's actually using them besides the `connpool` package. Also, they're only used once (within the `New()` function). Should we keep them exported?

oxtoacart: I don't have a strong feeling either way.  The reason that I typically export constants is that it can be a useful form of documentation.  And the reason I made these constants rather than inline was to describe what these numbers mean.

src/github.com/getlantern/connpool/connpool_test.go
```diff
@@ -94,38 +95,69 @@ func TestDialFailure(t *testing.T) {
...
+	claimTimeout := 5 * time.Second
+
+	cfg := &Config{
+		ClaimTimeout: claimTimeout,
+	}
+	p := New(*cfg)
+	defer p.Close()
+
+	cfg.ClaimTimeout = 7 * time.Second
+	assert.NotEqual(t, claimTimeout, cfg.ClaimTimeout, "Property changed on config shouldn't be reflected in pool")
```
xiam: I fail to see how this assertion proves that `p`'s configuration was in fact not affected, as we are only comparing `claimTimeout` and `cfg`. `p` must be not affected as we're using a copy of Config, but if we wanted to test it I think that casting `p` to `*pool` and checking if the initial value of `ClaimTimeout` is still equal to `claimTimeout` would give better results:

```go
assert.Equal(t, claimTimeout, p.(*pool).ClaimTimeout, "Property changed on config shouldn't be reflected in pool")
```

or using `cfg.ClaimTimeout` instead of `claimTimeout`:

```go
assert.NotEqual(t, cfg.ClaimTimeout, p.(*pool).ClaimTimeout, "Property changed on config shouldn't be reflected in pool")
```

oxtoacart: Good point, fixed.

src/github.com/getlantern/connpool/connpool_test.go
```diff
@@ -94,38 +95,69 @@ func TestDialFailure(t *testing.T) {
...
 }
 
-func connectAndRead(t *testing.T, p *Pool, loops int) {
+func connectAndRead(t *testing.T, p Pool, loops int) {
```
xiam: Works perfectly! I'd just like to make a quick suggestion, a `Pool` must be prepared to deal with concurrent Get()s so we could probably rephrase this routine a bit to fire concurrent Get()s instead of sequential and still wait for them all to complete (such as in the original):

```
func connectAndRead(t *testing.T, p Pool, loops int) {
	var wg sync.WaitGroup

	for i := 0; i < loops; i++ {
		wg.Add(1)

		func(wg *sync.WaitGroup) {
			c, err := p.Get()
			if err != nil {
				t.Fatalf("Error getting connection: %s", err)
			}
			read, err := ioutil.ReadAll(c)
			if err != nil {
				t.Fatalf("Error reading from connection: %s", err)
			}
			assert.Equal(t, msg, read, "Should have received %s from server", string(msg))
			c.Close()

			wg.Done()
		}(&wg)
	}

	wg.Wait()
}
```

oxtoacart: Great idea!  Done.

src/github.com/getlantern/enproxy/conn_test.go
```diff
@@ -62,18 +60,80 @@ func TestBadBuffered(t *testing.T) {
...
+
 	conn, err := prepareConn(httpAddr, buffered, false, t)
 	if err != nil {
 		t.Fatalf("Unable to prepareConn: %s", err)
 	}
-	defer conn.Close()
+	defer func() {
+		err := conn.Close()
+		assert.Nil(t, err, "Closing conn should succeed")
+		assert.NoError(t, counter.AssertDelta(2), "All file descriptors except the connection from proxy to destination site should have been closed")
```
xiam: Seems looks like we have an odd race condition here but I haven't figure out the cause. I found it while running `go tests` for the first time, but it didn't happend the second time until a ran `go test` inside a while loop.

```
--- FAIL: TestPlainTextStreaming (1.10s)
        Location:       conn_test.go:131
        Error:          No error is expected but got Unexpected TCP file descriptor count. Expected 2, have 0.

                        New file descriptors
                        -----------------------------
                        enproxy.t 22767  rev   10u     IPv4 511988      0t0     TCP localhost.localdomain:50181->localhost.localdomain:58955 (ESTABLISHED)
                        enproxy.t 22767  rev    9u     IPv4 511987      0t0     TCP localhost.localdomain:58955->localhost.localdomain:50181 (ESTABLISHED)

                        Removed file descriptors
                        -----------------------------
                        enproxy.t 22767  rev    7u     IPv4 512067      0t0     TCP localhost.localdomain:43616->localhost.localdomain:57883 (CLOSE_WAIT)
                        enproxy.t 22767  rev    9u     IPv4 510929      0t0     TCP localhost.localdomain:39772->localhost.localdomain:46442 (CLOSE_WAIT)
        Messages:       All file descriptors except the connection from proxy to destination site should have been closed
```

 The `CLOSE_WAIT` status may have something to do with it? or a waiting time somewhere? I'll investigate more on the topic.

I'm triggering the race condition and capturing the output with:

```
while true; do go test &>> log.txt; sleep 1; done
```

It eventually happens.

Any clues where should I start looking?

oxtoacart: Very interesting.  I just tried to reproduce this by running your script, and after running for 30 minutes, I didn't get any failures.  I did add some code that prints out the [connection trace](https://github.com/getlantern/flashlight-build/tree/peerscanner/src/github.com/getlantern/enproxy#debugging) information if the file descriptor count is wrong, so that might help you debug.

BTW, I'm running go 1.4.1 on OS X 10.9.5.

What's interesting about your output is that the new and removed sockets don't have any ports in common.  This makes me wonder whether we're seeing spillover from another test.  You could try commenting out all of the tests except the one that's failing and retrying that.

xiam: I ran this on Linux, I'll try again on a Mac to try to determina if this happens only under specific circumstances.

xiam: Hello @aranhoide, does this error look familiar to you?

I've got the same error on Mac (go1.4.1) so it could not be a Linux-only problem.

```
2015/01/30 10:51:27 http: TLS handshake error from 127.0.0.1:53729: EOF
enproxy: 
---- Enproxy Connections----
Open:                         0
Closing:                      0
Blocked on Closing:           0
Blocked on Read:              0
Reading:                      0
Reading Finishing:            0
Blocked on Write:             0
Writing:                      0
  Selecting:                  0
  Writing:                    0
    Write Pipe Open:          0
    Request Pending:          0
      Submitting Req.:        0
      Processing Req.:        0
        Posting Req. Fin:     0
        Posting Resp:         0       
        Dialing First:        0
        Redialing:            0
    Doing Write:              0
  Posting Response:           0
  Writing Empty:              0
  Finishing Body:             0
  Finishing:                  0
Requesting:                   0
Requesting Finishing:         0

--- FAIL: TestPlainTextStreaming (0.75s)
        Location:       conn_test.go:131
        Error:          No error is expected but got Unexpected TCP file descriptor count. Expected 2, have 1.

                        Removed file descriptors
                        -----------------------------
                        enproxy.t 1155 reventlov   10u    IPv4 0x6510f8cd1abdeec5       0t0     TCP localhost:53730->localhost:53731 (CLOSE_WAIT)

                        New file descriptors
                        -----------------------------
                        enproxy.t 1155 reventlov   12u    IPv4 0x6510f8cd1f1a8b85       0t0     TCP localhost:53733->localhost:53719 (ESTABLISHED)
                        enproxy.t 1155 reventlov   13u    IPv4 0x6510f8cd1f1bdd25       0t0     TCP localhost:53719->localhost:53733 (ESTABLISHED)
        Messages:       All file descriptors except the connection from proxy to destination site should have been closed
```

I'll continue studying the code, it would be great training fixing this.

xiam: So far I've found that the test doest not fail when we have these matching file descriptors:

```
startingLines: enproxy.t 2867 reventlov    5u    IPv4 0x6510f8cd1e78ed25       0t0     TCP localhost:59779 (LISTEN)
enproxy.t 2867 reventlov    8u    IPv4 0x6510f8cd1fb132b5       0t0     TCP localhost:59790 (LISTEN)
enproxy.t 2867 reventlov    9u    IPv4 0x6510f8cd1f28d795       0t0     TCP localhost:59792 (LISTEN)
```

But, sometimes a CLOSE_WAIT appears:

```
enproxy.t 2914 reventlov    5u    IPv4 0x6510f8cd1ce7cb85       0t0     TCP localhost:60430 (LISTEN)
enproxy.t 2914 reventlov    8u    IPv4 0x6510f8cd1efbaec5       0t0     TCP localhost:60456 (LISTEN)
enproxy.t 2914 reventlov    9u    IPv4 0x6510f8cd1f2f1795       0t0     TCP localhost:60458 (LISTEN)
enproxy.t 2914 reventlov   10u    IPv4 0x6510f8cd1fae65f5       0t0     TCP localhost:60458->localhost:60459 (CLOSE_WAIT)
```

and the test fails.

The CLOSE_WAIT fd is a temporary status, in fact, if I add a `time.Sleep(time.Second * 1)` after `startServers()` on the `doTestPlainText()` function the test does not fail anymore so my guess is that this is something that can be fixed within `startServers()`, which I'm now reviewing.

xiam: I've tracked down the mysterious opened-then-closed connection that caused our `CLOSE_WAIT` to the `WaitForServer()` function within the `waitforserver` package. The reason is this function attempts to dial to a network address and, if successful, it immediately quits. So the `CLOSE_WAIT` origin is now clear (and it's also completely expected).

The `WaitForServer()` function is mostly used in tests with the exception of `./github.com/getlantern/yamlconf/yamlconf_http.go`. I'd suggest to change the function a bit to allow the `CLOSE_WAIT` status to disappear. I suggest changing the `WaitForServer()` function directly since this change would be practically inoffensive and would avoid changing other parts of our code to add a timer.

```
func WaitForServer(protocol string, addr string, limit time.Duration) error {
  defer time.Sleep(50 * time.Millisecond)

  cutoff := time.Now().Add(limit)
  for {
    if time.Now().After(cutoff) {
      return fmt.Errorf("Server never came up at %s address %s", protocol, addr)
    }   
    c, err := net.DialTimeout(protocol, addr, limit)
    if err == nil {
      return c.Close()
    }   
    time.Sleep(50 * time.Millisecond)
  }
}
```

Another solution would be adding a `time.Sleep(time.Millisecond*100)` after the `startServers()` or after any occurrence of the `WaitForServer()` function.

src/github.com/getlantern/connpool/connpool.go
aranhoide: +1 on making these constants rather than using magic numbers.  -1 on exporting names unless we think external libraries should be able to use them.

src/github.com/getlantern/enproxy/conn_test.go
aranhoide: Great catch, @xiam!  I was indeed not familiar at all with this error.  So IIUC, `Close` "starts" closing the connection, but we are not guaranteed that the file descriptor is cleaned up when it returns?  And we believe that it definitely gets cleaned up within 50 milliseconds of closing?

If so, I'd just put the wait right at the beginning of `fdcount.Counter.AssertDelta`, which is where that matters.  Thus, that consideration would not pollute production code.  Also, it would hopefully correct for other cases of functions that create short-lived connections internally.

xiam: Thanks @aranhoide, in fact we can't be sure it will get cleaned on 50ms, specially is the testing server is under big load.

The problem with adding a wait time before AssertDelta would be that we're querying a file descriptor list [a bit earlier](https://github.com/getlantern/flashlight-build/blob/2673dcf11ae9383f2bd24640af5948534041fa61/src/github.com/getlantern/enproxy/conn_test.go#L119), and this early query is the one that has the CLOSE_WAIT inside.

I guess we could also ignore file descriptors in CLOSE_WAIT status [when matching](https://github.com/getlantern/flashlight-build/blob/2673dcf11ae9383f2bd24640af5948534041fa61/src/github.com/getlantern/fdcount/fdcount.go#L55), that won't require a sleep time at all and won't pollute production code either since `fdcount` is just used for testing purposes and would also correct other cases that depend on `fdcount` matching file descriptors, the only problem I see with this alternative is that having an abnormal number of CLOSE_WAIT file descriptors after a reasonable time could also indicate a problem.

aranhoide: Oh, I see.  What about something like

```go
func (c *Counter) matchingLines(out []byte) (string, int) {
	for try = 0; try < 10; try++ {
		s, i, anyCloseWait := c.matchingLinesInner(out)
		if anyCloseWait {
			time.Sleep((50 << try) * time.Millisecond)
		} else {
			return s, i
		}
	}
	// OK, it seems these pesky CLOSE_WAITs just won't go.  Let the calling
	// test handle this as it sees fit.
	s, i, _ = c.matchingLinesInner(out)
	return s, i
}

func (c *Counter) matchingLinesInner(out []byte) (string, int, bool) {
	lines := make([]string, 0)
	anyCloseWait := false
	for _, line := range strings.Split(string(out), "\n") {
		if strings.HasSuffix(line, "(CLOSE_WAIT)") {
			anyCloseWait := true
		}
		if strings.Contains(line, c.match) {
			lines = append(lines, line)
		}
	}
	return strings.Join(lines, "\n"), len(lines), anyCloseWait
}
```
It's ugly, but it's test code so... :P

Note, however, that this is totally untested and only meant to illustrate the concept: wait until no CLOSE_WAITs left, giving up after a good while (one second and change, unless I borked the arithmetic :) ).

aranhoide: Or maybe we only want to match a CLOSE_WAIT if it is matched and would thus be returned?  I have no idea whether this matters, but I guess it can't hurt.  So, just put the check for (CLOSE_WAIT) inside the `if` for the other one:
```go
func (c *Counter) matchingLinesInner(out []byte) (string, int, bool) {
    lines := make([]string, 0)
    anyCloseWait := false
    for _, line := range strings.Split(string(out), "\n") {
        if strings.Contains(line, c.match) {
            if strings.HasSuffix(line, "(CLOSE_WAIT)") {
                anyCloseWait := true
            }
            lines = append(lines, line)
        }
    }
    return strings.Join(lines, "\n"), len(lines), anyCloseWait
}
```

src/github.com/getlantern/peerscanner/host.go
```diff
@@ -0,0 +1,407 @@
...
+		statusCh:     make(chan chan *status, 1000),
+		proxiedClient: &http.Client{
+			Transport: &http.Transport{
+				Dial: func(network, addr string) (net.Conn, error) {
+					return enproxy.Dial(addr, &enproxy.Config{
+						DialProxy: func(addr string) (net.Conn, error) {
+							return tlsdialer.DialWithDialer(&net.Dialer{
+								Timeout: dialTimeout,
+							}, "tcp", ip+":443", true, &tls.Config{
+								InsecureSkipVerify: true,
```
xiam: Is this `InsecureSkipVerify` safe?

src/github.com/getlantern/peerscanner/main.go
```diff
@@ -0,0 +1,207 @@
...
+	defer hostsMutex.Unlock()
+	return hostsByIp[ip]
+}
+
+func isPeer(name string) bool {
+	// We just check the length of the subdomain here, which is the unique
+	// peer GUID. While it's possible something else could have a subdomain
+	// this long, it's unlikely.
+	// We also accept anything with a name beginning with peer- as a peer
+	return len(name) == 32 || strings.Index(name, "peer-") == 0
```
xiam: `strings.Index(name, "peer-") == 0` works but we could also use `strings.HasPrefix("name", "peer-")`.

src/github.com/getlantern/peerscanner/cf/cf.go
```diff
@@ -0,0 +1,82 @@
...
+			},
+		},
+	}
+	return &Util{client, domain}
+}
+
+func (util *Util) GetAllRecords() ([]cloudflare.Record, error) {
+	resp, err := util.Client.LoadAll(util.domain)
+	if err != nil {
+		return nil, fmt.Errorf("Error retrieving records!", err)
```
xiam: `fmt.Errorf("Error retrieving records!", err)` does not actually include the format code for displaying `err`, so it will get appended as `%!(EXTRA ...)`. We could use `fmt.Errorf("Error retrieving records!: %q", err)`.

src/github.com/getlantern/peerscanner/cf/cf.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+	allRecords := resp.Response.Recs.Records
+	for resp.Response.Recs.HasMore {
+		resp, err = util.Client.LoadAllAtIndex(util.domain, len(allRecords))
+		if err != nil {
+			return nil, fmt.Errorf("Error retrieving records at index!", err)
```
xiam: `fmt.Errorf("Error retrieving records at index!", err)` does not actually include the format code for displaying `err`, so it will get appended as `%!(EXTRA ...)`. We could fix it with `fmt.Errorf("Error retrieving records at index!: %q", err)`.

src/github.com/getlantern/peerscanner/host.go
oxtoacart: We don't care, we're just checking that there's connectivity to the proxy.  Also, as a general note, anyone can run one of these proxies, so there's nothing meaningful for us to verify anyway.

src/github.com/getlantern/peerscanner/main.go
```diff
@@ -0,0 +1,207 @@
...
+
+func getOrCreateHost(name string, ip string) *host {
+	hostsMutex.Lock()
+	defer hostsMutex.Unlock()
+
+	h := hostsByIp[ip]
+	if h == nil {
+		h := newHost(name, ip, nil)
+		hostsByIp[ip] = h
+		go h.run()
```
xiam: `h.run()` already calls `go h.doRun()` as a goroutine, maybe `go h.run()` is not strictly necessary.

src/github.com/getlantern/peerscanner/web.go
```diff
@@ -0,0 +1,150 @@
...
+			err = fmt.Errorf("Received invalid port for %v - %v: %v", name, ip, portString)
+		}
+	}
+
+	return
+}
+
+func clientIpFor(req *http.Request, name string) string {
+	// Client requested their info
+	clientIp := req.Header.Get("X-Peerscanner-Forwarded-For")
```
xiam: Third parties could send fake `X-Peerscanner-Forwarded-For` headers, and `clientIpFor()` will just trust the value the client is sending. Whatever this value is, it can be fed to `h := getOrCreateHost(name, ip)`. Should we worry about fake or badly formatted IPs, or about collisions with existent IPs (keys) on the `hostsByIp` map?

src/github.com/getlantern/peerscanner/web.go
```diff
@@ -0,0 +1,150 @@
...
+	if clientIp == "" {
+		clientIp = req.Header.Get("X-Forwarded-For")
```
xiam: Same case as above, third parties could send a fake `X-Forwarded-For` header and whatever this value is, it will be passed to `clientIp`.

src/github.com/getlantern/peerscanner/web.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+	if clientIp == "" && isFallback(name) {
+		// Use direct IP for fallbacks
+		clientIp = strings.Split(req.RemoteAddr, ":")[0]
```
xiam: If we put peerscanner behind a proxy (such as nginx), this `RemoteAddr` would be the IP of the proxy rather than the IP of the remote machine. This could be solved by accepting `X-Forwarded-For` or `X-Peerscanner-Forwarded-For` headers as above, but how can we trust or test those values?

The nginx web server has a (realip_module](http://nginx.org/en/docs/http/ngx_http_realip_module.html) that also trusts `X-Forwarded-For` headers, but only from a fixed whitelist. It don't think we could implement a whitelist since we don't know the servers that are going to try to register with peerscanner but maybe we could test them somehow before trusting the IP we're receiving.

src/github.com/getlantern/peerscanner/web.go
```diff
@@ -0,0 +1,150 @@
...
+	if err != nil {
+		resp.WriteHeader(http.StatusBadRequest)
+		fmt.Fprintln(resp, err.Error())
+		return
+	}
+
+	h := getHostByIp(ip)
+	msg := "Host not registered"
+	if h != nil {
+		h.unregister()
```
xiam: Seems like any third party that feeds a fake `X-Forwarded-For` header to `/unregister` would be able to trigger the unregister method of `hostsByIp[{x-forwarded-for}]` and affect other host.

src/github.com/getlantern/peerscanner/host.go
```diff
@@ -0,0 +1,407 @@
...
+// newHost creates a new host for the given name, ip and optional DNS record.
+func newHost(name string, ip string, record *cloudflare.Record) *host {
+	// Cache TLS sessions
+	clientSessionCache := tls.NewLRUClientSessionCache(1000)
+
+	h := &host{
+		name:         name,
+		ip:           ip,
+		record:       record,
+		resetCh:      make(chan string, 1000),
```
xiam: The `resetCh` and `statusCh` are large channel buffers inside a `host` struct that is mapped to a unique IP by means of the `hostsByIp` map. I think it is important to consider that, at this time, anyone is able to request `peerscanner` to allocate such struct (see `register()`) by sending an special `X-Forwarded-For` header.

.travis.yml
xiam: Hello @myleshorton, do you have a copy of these keys? I wanted to test `peerscanner/cf` and I guess the required env vars are here.

myleshorton: What keys did you use for this @oxtoacart?

src/github.com/getlantern/peerscanner/main.go
oxtoacart: Good call, fixed!

src/github.com/getlantern/peerscanner/cf/cf.go
oxtoacart: Fixed.

src/github.com/getlantern/peerscanner/cf/cf.go
oxtoacart: Fixed.

src/github.com/getlantern/peerscanner/main.go
oxtoacart: Fixed.

fffw:
Frankly, I still need effort to fully understand enproxy code, and didn't
notice the issue @xiam found during scaning peerscanner. Well done @xiam!
What I looked carefully was connpool which I believe is good enough now.

xiam:
@myleshorton Cool! let's do it then. The `enproxy` package does a lot of things, it works fine but I'll need to hack on it a bit more to actually master it, however the test looked solid and they ran with only a race condition that I can fix easily, so I think we'll have no problem with it. The other packages were smaller and easier to understand.

Just let me give another look at the whole PR and then I'll merge it into `devel`, I'll continue studying `enproxy` after the merge.

myleshorton:
Is the race condition new? If so I think we should fix it with this PR but either way should probably fix it soon.

src/github.com/getlantern/enproxy/conn_test.go
```diff
@@ -62,18 +60,82 @@ func TestBadBuffered(t *testing.T) {
...
+
 func doTestPlainText(buffered bool, t *testing.T) {
 	startServers(t)
 
+	_, counter, err := fdcount.Matching("TCP")
+	if err != nil {
+		t.Fatalf("Unable to get fdcount: %v", err)
+	}
+
 	conn, err := prepareConn(httpAddr, buffered, false, t)
```
xiam: I was looking for the comment I left here  about the race condition involving `CLOSE_WAIT` statuses but I can't find it anymore (probably was deleted accidentally). 

I recovered @aranhoide's suggestion from the copy that was emailed to us and I think his code would solve our problem without having to rely on timers:

```
func (c *Counter) matchingLines(out []byte) (string, int) {
    for try = 0; try < 10; try++ {
        s, i, anyCloseWait := c.matchingLinesInner(out)
        if anyCloseWait {
            time.Sleep((50 << try) * time.Millisecond)
        } else {
            return s, i
        }
    }
    // OK, it seems these pesky CLOSE_WAITs just won't go.  Let the calling
    // test handle this as it sees fit.
    s, i, _ = c.matchingLinesInner(out)
    return s, i
}

func (c *Counter) matchingLinesInner(out []byte) (string, int, bool) {
    lines := make([]string, 0)
    anyCloseWait := false
    for _, line := range strings.Split(string(out), "\n") {
        if strings.HasSuffix(line, "(CLOSE_WAIT)") {
            anyCloseWait := true
        }
        if strings.Contains(line, c.match) {
            lines = append(lines, line)
        }
    }
    return strings.Join(lines, "\n"), len(lines), anyCloseWait
}
```

The only problem I see with this is that a lot of long-lived `CLOSE_WAIT` statuses could also indicate a problem (that's not the case), but that's not critical and it's probably outside the scope of this PR.

So, if you agree, I'll use @aranhoide's suggestion to close this issue.

xiam:
@myleshorton I think it's not new but it is also not critical. It's actually a bug inside test code that does not affect production but only the test's output. See: https://github.com/getlantern/flashlight-build/pull/34/files#r24025896

oxtoacart:
@xiam I liked the idea behind @aranhoide's suggestion, but I don't think we want anything specific to CLOSE_WAIT in fdcount since it's pretty generic.  I implemented his basic idea, but slightly differently - https://github.com/getlantern/flashlight-build/commit/d03171ec63fae95abe053dd32df7c113daccb743.

.travis.yml
xiam: Thanks @oxtoacart!

@myleshorton I think that link ultimately refers to a private repository (I'm getting 404 while trying to access envvars.bash), could you add me there?

myleshorton: I just added it to too-few-secrets and will update the doc!

oxtoacart:
@aranhoide 

> Yes, AssertDelta is too late to help with CLOSE_WAITs in the original reading. That is the case in the bug that @xiam found.

Yes, xiam and I have addressed this now I think. See https://github.com/getlantern/flashlight-build/commit/8c38e8f5bafdb728155945c876183867c6a5619f.

> Also, "checking until we get what we expect" seems prone to overoptimistic test results. I'd rather sacrifice some of the genericity of fdcount and make it explicit that we consider CLOSE_WAIT a temporary and unpredictable condition that we are explicitly trying to take out of consideration.

Well, if the hypothesis is that CLOSE_WAIT can take a little time to clear, then we need to account for that at the end as well as the start.  My original change does that.  The way that we use this, we're always looking for the count to get down to some number.  Either it does or it doesn't, and waiting a little extra time for it to do so doesn't hurt anything.

xiam:
> Cool. Does this mean we can merge?

@oxtoacart Yeah! thanks a lot for your help. A lot of code was introduced today, please allow me a bit more time to review the new parts, just to make sure we didn't introduce any new issues.

oxtoacart:
Thanks for the thoroughness @xiam.

oxtoacart:
Thanks @xiam @fffw @myleshorton and @aranhoide for the review!

xiam:
@oxtoacart My pleasure! It was a great learning experience.

.travis.yml
aranhoide: I have changed these vars to use a CFL_ prefix, as part of a refactoring to avoid confusion with cloudfront.  How do I reencrypt the envvars.bash to produce a new envvars.bash.enc?  Where do I get the required key(s)?  I couldn't find anything travis-related in too-many-secrets.

aranhoide: @oxtoacart  ^

oxtoacart: Travis uses GitHub for authentication, so you should be able to log in with your own credentials per these instructions - http://docs.travis-ci.com/user/getting-started/#Step-one%3A-Sign-in

Once you're logged in, you can follow these instructions (linked from the README) to encrypt the file - http://docs.travis-ci.com/user/encrypting-files/

PR #35, opened 2015-01-24
Issue 2185

oxtoacart:
This adds logic to turn connpool into a lazy connpool (i.e. it only fills up when connections are actively being requested).  This gives most of the benefit without the drawback of continually opening connections while the client is idle.


oxtoacart:
@myleshorton FYI - these are some tweaks that I'm making along with the peerscanner changes to make connection pooling more robust.  It doesn't actually fix #2185 since that wasn't actually caused by connection pooling, but it's a useful change anyway.

PR #36, opened 2015-01-27
Lantern android

xiam:
The [lantern-android](https://github.com/getlantern/lantern-android) project provides tools for creating a flashlight library that makes easy running and controlling a flashlight client on Android devices.

This commit provides example code and instructions.


oxtoacart:
@xiam I'm starting to review this now.  I may not be able to finish before I leave on vacation, but I'll start taking notes.

This is an awesome PR!  The README is excellent.

src/github.com/getlantern/lantern-android/README.md
```diff
@@ -0,0 +1,232 @@
...
+      </intent-filter>
+    </activity>
+  </application>
+  <uses-permission android:name="android.permission.INTERNET" />
+</manifest>
+```
+
+Now add a new *main activity* by right-clicking on the top most directory on
+the *Project* pane and selecting New->Activity->Blank Activity, the default
+values would be OK, click *Finish*.
```
oxtoacart, line 97: @xiam This option is grayed out for me and Android studio complains that I need to set the minSdkVersion to at least 7.

If I update the defaultconfig in app/build.gradle to look like the below, then I can create an Activity:

``` groovy
    defaultConfig {
        applicationId "org.getlantern.example"
        minSdkVersion 7
    }
```

xiam:
@oxtoacart Thanks! actually the README on the main repo was a bit outdated compared to the copy I prepared for this PR but I just updated it :-).

src/github.com/getlantern/lantern-android/README.md
xiam, line 97: @oxtoacart I'm not sure what's causing that but maybe I can find a fix if I could replicate it, what's your Android Studio version? mine is this:

untitled

src/github.com/getlantern/lantern-android/README.md
```diff
@@ -0,0 +1,232 @@
...
+
+make
+# ...
+# BUILD SUCCESSFUL
+# Total time: 7 seconds
+```
+
+The `make` command will create a new `app` subdirectory that will contain an
+Android example project. You may import the contents of the `app` subdirectory
+into Android Studio to see libflashlight working.
```
oxtoacart, line 66: @xiam It turns out that I can't run the Android emulator on my Mac if VirtualBox is open.  I forget whether I opened VirtualBox myself or if boot2docker did it during the make process, but it might be nice to add a note about that in here.

xiam, line 66: @oxtoacart  You're right, I'll also need to add a note specifying that the ARM emulator is required, the default is the x86 emulator.

src/github.com/getlantern/lantern-android/README.md
oxtoacart, line 97: Weird, mine is older than yours, even though I just updated it ...

screen shot 2015-01-27 at 5 54 38 pm

oxtoacart, line 97: Hmm, I just checked again and an update is available.  I installed it, and now I'm at this:

screen shot 2015-01-27 at 5 56 58 pm

src/github.com/getlantern/lantern-android/README.md
```diff
@@ -0,0 +1,232 @@
...
+
+![Deploy to a device](https://cloud.githubusercontent.com/assets/385670/5712899/db6ddb34-9a7b-11e4-8841-6b6b12e46c27.png)
+
+As long as the app is open, you'll be able to test the canonical example by
+finding the device's IP and sending it a special request:
+
+```
+curl -x 10.10.100.97:9192 http://www.google.com/humans.txt
+# Google is built by a large team of engineers, designers, researchers, robots, and others in many different sites across the globe. It is updated continuously, and built with more tools and technologies than we can shake a stick at. If you'd like to help us out, see google.com/careers.
+```
```
oxtoacart, line 232: I'm not able to make this work, as I don't have access to the emulator's network.

What I can do however is set up redirection by first telnetting to the emulator's console and then setting up a redirect, like this:

``` bash
➜  ox.to.a.cart git:(master) ✗ telnet localhost 5556
Trying ::1...
telnet: connect to address ::1: Connection refused
Trying 127.0.0.1...
Connected to localhost.
Escape character is '^]'.
Android Console: type 'help' for a list of commands
OK
redir add tcp:9192:9192
OK
```

Then I can

```
curl -x localhost:9192 http://www.google.com/humans.txt
```

src/github.com/getlantern/lantern-android/client/client.go
```diff
@@ -0,0 +1,209 @@
...
+	rpInitialized bool
+
+	balInitialized bool
+	balCh          chan *balancer.Balancer
+}
+
+// AddFrontedServer adds a fronted server to the list.
+func (client *Client) AddFrontedServer(fs *frontedServer) error {
+	client.frontedServers = append(client.frontedServers, fs)
+	// TODO: Find the best way to add this server to the balancer list.
```
oxtoacart: @xiam Balancer was meant to be initialized through an idempotent operation (i.e. supplying a full list of dialers).  If the list ever changes, you can just stop the old balancer and initialize a new balancer using the new full list.  I've deliberately followed this pattern in a variety of places to reduce the complexity of individual adds and removes cascading down through the layers.

src/github.com/getlantern/lantern-android/client/client.go
```diff
@@ -0,0 +1,209 @@
...
+
+	rpCh          chan *httputil.ReverseProxy
+	rpInitialized bool
+
+	balInitialized bool
+	balCh          chan *balancer.Balancer
+}
+
+// AddFrontedServer adds a fronted server to the list.
+func (client *Client) AddFrontedServer(fs *frontedServer) error {
```
oxtoacart: Why is this exported?  Is this actually a public API?

src/github.com/getlantern/lantern-android/client/client_test.go
```diff
@@ -0,0 +1,108 @@
...
+	"testing"
+	"time"
+)
+
+const listenProxyAddr = "127.0.0.1:9997"
+
+var testURLs = map[string][]byte{
+	"http://www.google.com/humans.txt": []byte("Google is built by a large team of engineers, designers, researchers, robots, and others in many different sites across the globe. It is updated continuously, and built with more tools and technologies than we can shake a stick at. If you'd like to help us out, see google.com/careers.\n"),
+	// TODO: This is not working, we actually need to implement a CONNECT request to proxy HTTPs traffic.
+	// "https://www.google.com/humans.txt": []byte("Google is built by a large team of engineers, designers, researchers, robots, and others in many different sites across the globe. It is updated continuously, and built with more tools and technologies than we can shake a stick at. If you'd like to help us out, see google.com/careers.\n"),
```
oxtoacart: @xiam You should be able to configure the http.Transport with a proxy, and it will take care of doing the CONNECT request for you.

src/github.com/getlantern/lantern-android/client/client_test.go
```diff
@@ -0,0 +1,108 @@
...
+	fmt.Printf(string(buf))
+
+	if bytes.Equal(buf, expectedContent) == false {
+		return fmt.Errorf("The response we've got from %s differs from what we expected.", uri)
+	}
+
+	return nil
+}
+
+func TestListenAndServeStop(t *testing.T) {
```
oxtoacart: What is this testing?  It doesn't have any assertions.

src/github.com/getlantern/lantern-android/client/client.go
xiam: The `Client` type must be exported so it can be used in the library bindings: https://github.com/getlantern/lantern-android/blob/master/libflashlight/bindings/main.go, but I see your point and certainly other types this package is exporting such as `CA`, `Config` and `Listener` aren't mean to be used outside the client so I'd better fix them.

src/github.com/getlantern/lantern-android/client/config.go
```diff
@@ -0,0 +1,122 @@
...
+)
+
+// pullConfigFile attempts to retrieve a configuration file over the network,
+// then it decompresses it and returns the file's raw bytes.
+func pullConfigFile() ([]byte, error) {
+	var err error
+	var res *http.Response
+
+	// Issuing a post request to download configuration file.
+	if res, err = http.Get(remoteConfigURL); err != nil {
```
oxtoacart: This will rarely work in the censored world.  We have to assume that the site from which we're fetching the updated configuration is blocked.  In fact, the site we use right now is s3.amazonaws.com, and that is periodically blocked in China.  That's why flashlight uses proxy for the http client - https://github.com/getlantern/flashlight-build/blob/peerscanner/src/github.com/getlantern/flashlight/config/config.go#L290

What flashlight actually does is the following:
1. Start using a list of servers and masquerades that's compiled into the binary (I see that you have that list, but aren't actually starting with it unless you're unable to fetch the list from the cloud)
2. Periodically fetch updates from the cloud by making proxied http requests (I see that you're only checking once, which is okay for now)
3. Update its configuration whenever a change has been detected in the cloud
4. Save the configuration to disk for whenever it next starts up (it looks like you're not saving it anywhere?)

The idea behind this is that the binary that we distribute will work for a while, but we may have to change our set of masquerades and servers to stay ahead of the censors.  As long as we're actually staying ahead of the censors, this will keep any Lantern clients that are already running unblocked.  And since these clients save their configuration to disk, they should remain unblocked even after a restart.

Ultimately, I think we'll want to do something similar for the Android version, but we don't need to be that sophisticated yet.  I think the way that you have should be fine with two changes:
1. Immediately start proxying based on the built-in configuration
2. Check for the cloud configuration through the already running proxy

src/github.com/getlantern/lantern-android/client/conn.go
```diff
@@ -0,0 +1,63 @@
+package client
```
oxtoacart: Maybe this file should be called listener.go?

src/github.com/getlantern/lantern-android/client/client_test.go
xiam: I was used to create tests relying on the `testing` package only, do you think I could improve my tests by using `github.com/getlantern/testify/assert` and adding appropriate descriptions?

src/github.com/getlantern/lantern-android/client/conn.go
```diff
@@ -0,0 +1,63 @@
...
+
+// Accept returns the next connection to the listener.
+func (li *Listener) Accept() (net.Conn, error) {
+	for {
+		select {
+		case <-li.closed:
+			return nil, ErrClosed
+		default:
+			return li.TCPListener.Accept()
+		}
```
oxtoacart: I don't think this will work quite as intended.  In particular, TCPListener.Accept() will block until a new connection is made.  If no new connection is made, the loop will never get a chance to check li.closed.

Is there a reason you don't just use a regular listener and call its Close() method when you want to stop listening?

src/github.com/getlantern/lantern-android/client/conn.go
```diff
@@ -0,0 +1,63 @@
...
+var (
+	// ErrCouldNotCreateListener is returned when NewListener fails.
+	ErrCouldNotCreateListener = errors.New(`Could not create new listener.`)
+	// ErrClosed is returned when the server is closed.
+	ErrClosed = errors.New(`Server was manually closed.`)
+)
+
+// Listener is a wrapper around net.TCPListener that attempts to provide a
+// Stop() function.
+type Listener struct {
```
oxtoacart: As a rule of thumb, I prefer to not export anything unless it needs to be used outside of its package.  Listener seems like it's completely internal to the client package.

oxtoacart:
@xiam Very nice!  Other than my inline comments, the only overarching thing that I see is that this ends up duplicating a fair amount of the code in flashlight.  I don't think we need to resolve this for the current PR, but I think it might make sense to create a flashlight lib that does all of the core stuff that's common between flashlight and lantern-mobile (e.g. configuration, bootstrapping fronted, etc.).  Doing that well will require some thought, so there's no need to hurry it.

src/github.com/getlantern/lantern-android/client/client.go
oxtoacart: Makes sense. My question was actually about the `AddFrontedServer` method specifically.  Is that something that's part of the public API here?

src/github.com/getlantern/lantern-android/client/client_test.go
oxtoacart: I find the stuff written using assert to be easier to read and debug, but I don't think it's required.

My question here was actually about the `TestListenAndServeStop` method.  It calls some APIs, but it doesn't actually check anything (e.g. return values, ability to connect/proxy via a port, file descriptor counts, etc.), so it's not clear to me that it's actually testing anything.

src/github.com/getlantern/lantern-android/client/config.go
xiam: @oxtoacart Yeah, you're right, we can't suppose we're going to have this URL always available.

> Immediately start proxying based on the built-in configuration
> Check for the cloud configuration through the already running proxy

This sounds good, I could give it another try.

> Save the configuration to disk for whenever it next starts up (it looks like you're not saving it anywhere?)

We have a limitation on Android and is that writing files is not that easy unless we have a writable path, as far as I know, this path is different for each application namespace, and we don't have that information unless the app provides it to us. We could export a new function that the app could use to set this path and we could save the configuration file, but I'm not sure if this would be the best solution on the Android world.

oxtoacart: > We have a limitation on Android and is that writing files is not that easy unless we have a writable path, as far as I know, this path is different for each application namespace, and we don't have that information unless the app provides it to us. We could export a new function that the app could use to set this path and we could save the configuration file, but I'm not sure if this would be the best solution on the Android world.

Yeah.  It seems that in Android, it might be better to have the app itself manage the configuration and just pass it in to the Go side (perhaps just as one big config object).  Anytime the app wants to change the configuration, it can just pass in the whole new configuration.  This would actually be very similar to how flashlight works - anytime that any part of the configuration is updated, the whole configuration is propagated through the system*.  Individual components are free to ignore it if their part of it hasn't changed.
- \- In reality, only portions of the configuration are propagated, based on what each individual component needs/has access to.

xiam:
@oxtoacart Thanks for taking the time to let me some comments before your vacation! Allow me some time to clean mi pull request.

> but I think it might make sense to create a flashlight lib that does all of the core stuff

Yes, I ended up copying/duplicating lots of functionality to get through. Creating a flashlight library would be great! I think it is not that difficult but it won't be straightforward, we need to reduce dependencies and think of the limitations we have on Android too, so yes, we need to discuss it and give it some thought.

oxtoacart:
> Thanks for taking the time to let me some comments before your vacation! Allow me some time to clean mi pull request.

You're welcome.

src/github.com/getlantern/lantern-android/client/client.go
xiam: Thanks for explaining @oxtoacart, then we may not need this. I think we're going to use one client on Android so I'll drop this.

src/github.com/getlantern/lantern-android/client/client.go
xiam: No, it's not. I made it public before actually using the config file, but it really does not make sense now so I'll drop to too.

src/github.com/getlantern/lantern-android/client/client_test.go
xiam: After reading some of your tests on the code review I think I've learnt a better way to distribute my own tests, I'm sure this will be better after a bit of refactoring.

src/github.com/getlantern/lantern-android/client/client_test.go
xiam: I initially though this configuration was correct, but I can't make it work for HTTPs proxying.

``` go
        Transport: &http.Transport{
            Dial: func(n, a string) (net.Conn, error) {
                return net.Dial("tcp", listenProxyAddr)
            },
        },
```

Can you spot what I'm missing or point me to another similar test I could read?

Thanks!

atavism: Hi @xiam, since @oxtoacart is on vacation, I'll chime in, but he can correct me if I happen to be mistaken. You'll want to configure http.Transport with a Proxy: 

``` go
    client := &http.Client{
        Transport: &http.Transport{
            Proxy: func(req *http.Request) (*url.URL, error) {
                return url.Parse(listenProxyAddr)
            },
            Dial: func(n, a string) (net.Conn, error) {
                return net.Dial("tcp", listenProxyAddr)
            },
        },
    }
```

A similar example can be found in flashlight_test.go

xiam: Cool! now I understand, thank you very much @atavism!

src/github.com/getlantern/lantern-android/client/conn.go
xiam: > Is there a reason you don't just use a regular listener and call its Close() method when you want to stop listening?

No, it was a stupid mistake. I was actually trying to create an error within the `http.Server.Serve()` method, specifically on this line: http://golang.org/src/net/http/server.go#L1728 that stopped the http server, so I thought that I could wrap `li.Accept()` to make it produce an error when I wanted. But `Close()` produced the error I needed already. I could swear I made a test before with a plain `net.Listener` that didn't work and kept blocking instead.

Sorry, I don't know what was I thinking but it had to be something rather dumb. I removed this thing completely from the source (even though it will remain in my memory for a while, haha).

Thank you very much for pointing this out @oxtoacart :-).

src/github.com/getlantern/lantern-android/README.md
xiam, line 66: I could not reproduce this error:

screen shot 2015-01-30 at 6 31 42 pm

does this happen every time you try to run the Android emulator when VirtualBox is already opened?

src/github.com/getlantern/lantern-android/client/config.go
xiam: @oxtoacart (I'm just replying to advance on this issue before I took another task, but this is not a top priority :-))

I just pushed some changes that will enable pulling the cloud configuration file over proxies periodically.

This is currently implemented like this:
- When the library starts, it immediately attempts to pull a fresh configuration directly, if something fails (timed out, access denied by censors, untrusted ssl connection, etc) it will continue with the default configuration.
- When a proxy starts (this will be requested by the Android app), it will fire a goroutine that will periodically check for a new configuration file, this pull will be through the client proxy itself. If any change is detected (based on the `Etag` and `If-None-Match` headers), the client will reload its configuration the same way the original flashlight client does.

xiam:
I'm closing this to target our current development branch (valencia) instead.

xiam:
See: https://github.com/getlantern/flashlight-build/pull/139

PR #37, opened 2015-01-27
getlantern/lantern#2196 Added support for MinQOS configuration

oxtoacart:
This is a fix for issue 2196, built off the peerscanner branch in anticipation of that

### DIFF
--- a/src/github.com/getlantern/flashlight/client/handler.go
+++ b/src/github.com/getlantern/flashlight/client/handler.go
@@ -43,7 +43,7 @@ func (client *Client) intercept(resp http.ResponseWriter, req *http.Request) {
 	addr := hostIncludingPort(req, 443)
 
 	// Establish outbound connection
-	connOut, err := client.getBalancer().DialQOS("tcp", addr, targetQOS(req))
+	connOut, err := client.getBalancer().DialQOS("tcp", addr, client.targetQOS(req))
 	if err != nil {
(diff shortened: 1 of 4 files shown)


src/github.com/getlantern/flashlight/client/handler.go
```diff
@@ -43,7 +43,7 @@ func (client *Client) intercept(resp http.ResponseWriter, req *http.Request) {
 	addr := hostIncludingPort(req, 443)
 
 	// Establish outbound connection
-	connOut, err := client.getBalancer().DialQOS("tcp", addr, targetQOS(req))
+	connOut, err := client.getBalancer().DialQOS("tcp", addr, client.targetQOS(req))
 	if err != nil {
 		respondBadGateway(clientConn, fmt.Sprintf("Unable to handle CONNECT request: %s", err))
 		return
```
myleshorton, line 49: Is it possible to get an err back here where connOut is not nil? I.e. is it possible there's a connection closing leak here because we're not calling something like:

``` go
if connOut != nil {
    defer connOut.Close()
}
```

oxtoacart, line 49: Conventionally, dial functions return either an error or the conn, but not both.  balancer follows that convention.

myleshorton, line 49: OK cool. LGTM otherwise.

oxtoacart:
Thanks!

PR #38, opened 2015-02-01
Multi front sticky routing

aranhoide:
This was tested with a server that is running at fl-ara-001.flashlightproxy.com.  To test the Cloudflare front, this was registered as fl-ara-001-orange.flashlightproxy.org in Cloudflare, and as d1xa31qki8z318.cloudfront.net in Cloudfront.  When testing this you probably need to set `insecureskipverify: true`.

Backwards compatibility with the `Host` configuration option was kept in enproxy and fronted, just in case anyone else was using those libraries, but not in flashlight itself.

I have manually tested this.  I couldn't think of any useful unit tests.  The main regression risk relates to stability of headers that CDN providers introduce in requests to flashlight servers (but I'm testing several headers per provider to try and mitigate against this risk).


aranhoide:
When testing this you may find the following flashlight.yaml useful to reach Cloudflare

``` yaml
version: 2
cloudconfig: ""
cloudconfigca: ""
addr: localhost:9999
role: client
instanceid: ""
country: xx
statsaddr: ""
cpuprofile: ""
memprofile: ""
waddellcert: ""
stats:
  reportingperiod: "0"
  statshubaddr: pure-journey-3547.herokuapp.com
server:
  unencrypted: false
  country: ""
  portmap: 0
  frontfqdns: {}
  waddelladdr: ""
client:
  minqos: 0
  dumpheaders: false
  frontedservers:
  - host: fl-ara-001-orange.flashlightproxy.org
    port: 443
    masqueradeset: cloudflare
    insecureskipverify: true
    bufferrequests: false
    dialtimeoutmillis: 0
    redialattempts: 2
    weight: 1000000
    qos: 10
  chainedservers: {}
  peers: {}
  masqueradesets:
    cloudflare:
    - domain: reddit.com
      ipaddress: 198.41.208.140
```

and the following one to reach Cloudfront (note a different port is used; for some reason I wanted to run these at the same time, so I had these `flashlight.yaml`s in two different folders; this probably doesn't matter)

``` yaml
version: 2
cloudconfig: ""
cloudconfigca: ""
addr: localhost:8888
role: client
instanceid: ""
country: xx
statsaddr: ""
cpuprofile: ""
memprofile: ""
waddellcert: ""
stats:
  reportingperiod: "0"
  statshubaddr: pure-journey-3547.herokuapp.com
server:
  unencrypted: false
  country: ""
  portmap: 0
  frontfqdns: {}
  waddelladdr: ""
client:
  minqos: 0
  dumpheaders: false
  frontedservers:
  - host: d1xa31qki8z318.cloudfront.net
    port: 443
    masqueradeset: cloudfront
    insecureskipverify: true
    bufferrequests: false
    dialtimeoutmillis: 0
    redialattempts: 2
    weight: 1000000
    qos: 10
  chainedservers: {}
  peers: {}
  masqueradesets:
    cloudfront:
    - domain: a0.awsstatic.com
      ipaddress: 54.230.79.95
```

To make sure the right headers were being returned, I temporarily added the following printouts to `enproxy/dialer.go:enproxyConfigWith`:

```
func (d *Dialer) enproxyConfigWith(dialProxy func(addr string) (net.Conn, error)) *enproxy.Config {
    return &enproxy.Config{
        DialProxy: dialProxy,
        NewRequest: func(upstreamHost string, method string, body io.Reader) (req *http.Request, err error) {
            if upstreamHost == "" {
                // No specific host requested, use configured one
                fmt.Println("No upstream host! Defaulting to ", d.Host)
                upstreamHost = d.Host
            } else {
                fmt.Println("Upstream host is: ", upstreamHost)
            }
            return http.NewRequest(method, "http://"+upstreamHost+"/", body)
        },
        BufferRequests: d.BufferRequests,
        IdleTimeout:    idleTimeout, // TODO: make this configurable
    }
}
```

aranhoide:
Finally, fl-ara-001.flashlightproxy.com is a regular salt-managed flashlight server that you can ssh into.   The modifications in that machine are:
- use a binary from this branch,
- use the `-nonencrypted` flag and listen in port 62000, and
- add the `-frontfqdns` flag.

This is how flashlight is invoked in `/etc/init/flashlight.conf`.  Note that the `-registerat` flag was removed, but the standard salt config does that when deploying to non-production setups.

```
exec /usr/bin/flashlight -addr :62000 -role server -unencrypted -frontfqdns '{cloudflare: fl-ara-001-orange.flashlightproxy.org, cloudfront: d1xa31qki8z318.cloudfront.net}' -instanceid "fl-ara-001" -statsperiod 300   2>&1 | logger -t flashlight
```

src/github.com/getlantern/enproxy/proxy.go
```diff
@@ -26,10 +26,16 @@ type Proxy struct {
...
+	// Host: FQDN of this particular proxy.  Either this or HostFn is required
+	// if this server was originally reached by DNS round robin.
 	Host string
 
+	// HostFn: given a http.Request, return the FQDN of this particular proxy,
+	// hopefully through the same front.  This is used to support multiple
+	// domain fronts.  Either this or Host is required if this server was
+	// originally reached by DNS round robin.
+	HostFn func(*http.Request) string
+
```
oxtoacart, line 38: I like the backward-compatibility of the API here, nice touch.

src/github.com/getlantern/flashlight/config/flags.go
```diff
@@ -89,3 +100,14 @@ func (updated *Config) applyFlags() error {
 
 	return nil
 }
+
+func parseFrontFQDNs(frontFQDNs string) (map[string]string, error) {
```
oxtoacart, line 106: I like the use of YAML here.

src/github.com/getlantern/flashlight/server/server.go
```diff
@@ -28,15 +28,37 @@ const (
...
+		// library normalizes all header names so the first letter of every
+		// dash-separated "word" is uppercase while all others are lowercase.
+		// Also, try and check more than one header to lean on the safe side.
+		"cloudflare": func(req *http.Request) bool {
+			return hasHeader(req, "Cf-Connecting-Ip") || hasHeader(req, "Cf-Ipcountry") || hasHeader(req, "Cf-Ray") || hasHeader(req, "Cf-Visitor")
+		},
+		"cloudfront": func(req *http.Request) bool {
+			return hasHeader(req, "X-Amz-Cf-Id") || headerMatches(req, "User-Agent", "Amazon Cloudfront")
+		},
+	}
```
oxtoacart, line 46: Rather than depending on headers that the CDNs may be setting, might it be safer to just have the client pass a header identifying which CDN it's using?  I know this means that old clients won't be able to use the new domain fronts, but I think that's okay.  This will also make things easier when we add additional CDNs, since we don't have to worry about whether or not they provide headers that are distinguishable from other CDNs.

src/github.com/getlantern/flashlight/server/server.go
```diff
@@ -28,15 +28,37 @@ const (
...
 )
 
+func headerMatches(req *http.Request, name string, value string) bool {
+	h := req.Header[name]
+	return len(h) == 1 && h[0] == value
```
oxtoacart: Does this match need to be so strict?

src/github.com/getlantern/flashlight/server/server.go
```diff
@@ -271,3 +293,26 @@ func onBytesGiven(destAddr string, req *http.Request, bytes int64) {
...
+		// We don't know about this provider... for backwards
+		// compatibility, let's try just any of the supplied FQDNs.
+		log.Debugf("Falling back to just any FQDN")
+		for _, fqdn := range fqdns {
+			return fqdn
+		}
+		// should never be reached, because the client is checking for empty
+		// `-frontfqdn`s, but Go doesn't know that and insists we must return
+		// something here.
+		return ""
```
oxtoacart, line 361: A less editorial comment might be something like `// No FQDN found` :)

oxtoacart, line 361: BTW, this condition will actually be reached if fqdns is empty, which from what I can tell is possible.

oxtoacart:
@aranhoide Looks good generally.  The big things are:
1. I think you need to reverse merge to ensure a clean forward merge (sorry it took me so long to review).
2. Please consider having the client pass a header that we use to identify the CDN rather than trying to identify it based on CDN-specific headers, primarily because we don't know if all future CDNs will be uniquely identifiable based on their own headers.
3. Please add a unit test to enproxy to test the use of HostFn (when you reverse merge, you'll see a new test condition around the old Host parameter, which you can probably mimic to test the HostFn parameter).  If you replace HostFn with a new header from the client, please test that instead.
4. If you stick with auto-identifying CDNs based on their own headers, please add a unit test to flashlight for this.  Maybe also consider moving the detection logic into its own library (cdndetect or something like that), since we will want to use it from lantern-android as well.  Like I said though, I think it may be best to just have the enproxy client set a new header that identifies the CDN or something like that.

After you've made any changes coming out of this, I'll go ahead and test locally.

src/github.com/getlantern/flashlight/config/flags.go
```diff
@@ -89,3 +100,14 @@ func (updated *Config) applyFlags() error {
...
+	fqdns := map[string]string{}
+	if err := yaml.Unmarshal([]byte(frontFQDNs), fqdns); err != nil {
+		return nil, err
+	}
+	if len(fqdns) == 0 {
```
aranhoide: @oxtoacart: we are checking here for an empty set of fqdns.  This error gets propagated up to the call to `config.Start` in `flashlight.go:main`, where it triggers `golog.Fatal`, exiting the program:

``` go
    if err != nil {
        log.Fatalf("Unable to start configuration: %s", err)
    }
```

That's why I believe we should never reach the condition below, where we have an empty mapping of fqdns.

src/github.com/getlantern/flashlight/server/server.go
aranhoide, line 361: @oxtoacart : I don't think it is possible since we'd have exited the program before ever starting to listen for client connections.  I've just added a github comment where I'm checking for this.  This is also why I added the editorial comment; if I believed this could be reached, some noisier error handling might be warranted here.

src/github.com/getlantern/flashlight/server/server.go
```diff
@@ -271,3 +293,26 @@ func onBytesGiven(destAddr string, req *http.Request, bytes int64) {
...
+				return fqdns[provider]
+			}
+		}
+		// We don't know about this provider... for backwards
+		// compatibility, let's try just any of the supplied FQDNs.
+		log.Debugf("Falling back to just any FQDN")
+		for _, fqdn := range fqdns {
+			return fqdn
+		}
+		// should never be reached, because the client is checking for empty
```
aranhoide: I meant the server.

src/github.com/getlantern/flashlight/config/flags.go
oxtoacart: Looks good, thanks!

oxtoacart: Actually, if one manually updates the yaml file, this validation won't apply.

aranhoide: True.  I'll try and catch it there too.   Another option is to change the signature of HostFn so it returns an error, but I was trying to avoid that because HostFn runs a bit more removed from any (server) user action and I saw no good way to handle errors there.

src/github.com/getlantern/flashlight/server/server.go
aranhoide, line 46: In this case I don't think we should sacrifice current functionality for fear of a hypothetical problem in the future.
- Most Lantern clients will be old for a while.  If we don't support old clients, blocking our current front would leave out the vast majority of our users.  I think if that was actually OK then adding new fronts wouldn't be a priority right now.
- I don't suppose we'll be adding any new fronts (except perhaps Azure, but it makes little sense to multifront that) before we have auto-updates.  After that, we can transition to the scheme you suggest very quickly if needed.
- In extreme cases we can still use fronts that we can't detect.  If we know a front is the only unblocked one we can remove all others from the `frontfqdns` and it will always be selected.

src/github.com/getlantern/flashlight/server/server.go
aranhoide: I don't think it matters in practice, esp. since we're additively checking several features of the request, but I'll iterate over `h` looking for a match.

src/github.com/getlantern/flashlight/server/server.go
aranhoide, line 46: So maybe the safest option would be that clients send this custom header to declare front, and the server uses if it present, falling back to checking CDN-specific headers if not.  I'm inclined to go YAGNI on this, since it seems so unlikely to ever make a difference, and there's some desirable simplicity in leaving the client out of this.

src/github.com/getlantern/flashlight/config/flags.go
oxtoacart: The other way to handle this would be to just allow an empty fqdns list and treat it the same way you already treat a nil list.

src/github.com/getlantern/flashlight/server/server.go
oxtoacart, line 46: > Most Lantern clients will be old for a while. If we don't support old clients, blocking our current front would leave out the vast majority of our users. I think if that was actually OK then adding new fronts wouldn't be a priority right now.

I wasn't suggesting that we block the current front.  The way that I would do it is to have a default front (the current) and use a different front if the new header is received.  That way, new clients can benefit from the additional fronts, while old clients can continue to operate as they have been.  Gaining access to additional fronts might be a good reason to upgrade your client, which is something that we want to encourage anyway.

The thing that doesn't feel right about the auto-detection functionality to me is that it relies on headers that weren't meant to be used for this purpose.  Even sticking with the current fronts, do we have any assurance that these headers are included 100% of the time and that they won't change without notice?  Or what if we end up tweaking some setting that affects the headers somehow?  The impact if the auto-detection fails is pretty severe, as we'll potentially be returning incorrect hosts which will basically make the service unusable.

> there's some desirable simplicity in leaving the client out of this

I actually think it would be simpler to use our own header rather than to reverse engineer the headers used by CDNs.  It's a technique that will work for all CDNs and overall seems less likely to break since we define and control the protocol.

All that said, I'm not putting my foot down, just making a suggestion - if you still think that autodetecting is the way to go, I won't stand in your way.

src/github.com/getlantern/flashlight/server/server.go
oxtoacart: Cool.

src/github.com/getlantern/flashlight/config/flags.go
oxtoacart: Actually, if you just take out the validation from here and leave the rest of the code as is (other than changing the comment in hostFn(), I think this will work nicely.

aranhoide: In the case of the command line, while a logical interpretation, in practice that looks like it would only hide user errors.  When loading from a flashlight.yaml it makes sense, since I imagine in that case the YAML reader will leave that field as the zero value (an empty dict), so that would be common.  So I think I'll do what you say: treat {} the same as nil, _but_ still keep the check here, so you don't accidentally enter an empty dict in the command line.  What do you think?

oxtoacart: I like it.

Just for the record, there's one more weird edge case.  Let's say that I previously configured the app with some fqdns via the command line - this will now be captured in the YAML.  If you allowed an empty list from the command line, I could use the command line to override what was in the YAML.  If you don't, I can't.  I can't actually think of a legitimate use case for this, so I wouldn't worry about it, but it's a weird nuance about how the configuration system works that I'm not 100% comfortable with and I wanted to make you aware of it.

src/github.com/getlantern/flashlight/server/server.go
aranhoide, line 46: I didn't understand that you were suggesting to block our current front.  I should have said "then if your current front was blocked [by censors] that would leave out the vast majority of users".  I think this is a problem and not a feature.  I don't think we'll have any trouble finding incentives for users to upgrade when it makes sense (e.g., when we get auto-updates out).  Limiting fronts doesn't seem like a particularly good one, since it doesn't affect the day to day experience until Lantern stops working altogether.  We might as well tell them that Lantern will stop working in X days.

TBH I think misdetection because these headers change is a theoretical rather than practical concern.  Although these headers may not be meant for this (and in some cases they seem to be meant for this, e.g., "User-Agent: Amazon Cloudfront"), they are meant for _something_ and stuff will break for millions of customers if these headers get changed willy-nilly.  Since we are checking different headers per CDN, we have some redundancy there; some headers could change and we'd still detect the right CDN.

I'll take a look at what would need to change in the client and if it seems like a small, safe change I'll add these headers.  I don't think they merit a significant or risky change at this point (at the very least, until the full-blown Lantern has auto-updates).

src/github.com/getlantern/flashlight/config/flags.go
aranhoide: Good catch!  Yeah, that's a good use case for an empty dict.  I'll remove this special handling and just let `-front-fqdn {}` clear the `FrontFQDNs`.

src/github.com/getlantern/flashlight/server/server.go
oxtoacart, line 46: > I didn't understand that you were suggesting to block our current front. I should have said "then if your current front was blocked [by censors] that would leave out the vast majority of users". I think this is a problem and not a feature. I don't think we'll have any trouble finding incentives for users to upgrade when it makes sense (e.g., when we get auto-updates out). Limiting fronts doesn't seem like a particularly good one, since it doesn't affect the day to day experience until Lantern stops working altogether. We might as well tell them that Lantern will stop working in X days.

That reminds me, I guess we'll need to update the yaml configurations for old clients to support the new fronts then.  We had only been publishing updates to the yaml configuration for the most recent client until now.

> TBH I think misdetection because these headers change is a theoretical rather than practical concern. Although these headers may not be meant for this (and in some cases they seem to be meant for this, e.g., "User-Agent: Amazon Cloudfront"), they are meant for something and stuff will break for millions of customers if these headers get changed willy-nilly. Since we are checking different headers per CDN, we have some redundancy there; some headers could change and we'd still detect the right CDN.

Likely true.

> I'll take a look at what would need to change in the client and if it seems like a small, safe change I'll add these headers. I don't think they merit a significant or risky change at this point (at the very least, until the full-blown Lantern has auto-updates).

Since you're keeping auto-detection in, I'd go with your instinct of leaving this out until we find a case where it's absolutely necessary.

aranhoide:
@oxtoacart : I made most of the changes you suggested.
- As per our discussion, I'm sticking to request inspection at least until we officially stop supporting old clients.  
- While I did a minimal test for the hostFn mechanism, I didn't do any automated tests for the detection of cloudflare and cloudfront headers.  That code won't change until we observe that CDN headers change in the field, so testing against captured headers would be pretty tautological.
- I didn't separate header inspection into its own library since that will only be used in Give mode and I don't suppose people will be giving access from their Android devices (correct me if I'm wrong).

oxtoacart:
@oxtoacart : I made most of the changes you suggested.

Sweet.
- As per our discussion, I'm sticking to request inspection at least until we officially stop supporting old clients.  

Makes sense.
- While I did a minimal test for the hostFn mechanism, I didn't do any automated tests for the detection of cloudflare and cloudfront headers.  That code won't change until we observe that CDN headers change in the field, so testing against captured headers would be pretty tautological.

Tautological yes, but also self documenting and a cheap way to increase our test coverage a little :)  It also has the benefit of providing regression testing so that if someone happens to mess around in that code, we'll know that we can still detect CloudFlare and cloudfront correctly.
- I didn't separate header inspection into its own library since that will only be used in Give mode and I don't suppose people will be giving access from their Android devices (correct me if I'm wrong).

Not unless we buy stock in some battery manufacturing companies :)

aranhoide:
I remain unconvinced that tests would be pulling their weight in this case.

I don't think in this case tests would document anything more than the code is documenting.  They would just restate what we believe about the field in not so different terms.

Test coverage is a proxy measurement for confidence vs regressions.  It's ~worthless to increase it per se if we know confidence is not substantially increased.  So I think regression testing would be the only compelling argument for some tests here.  And, as I said, I think the only reason to change that code would be if CDN headers changed.  So tests would need to be changed along with the code.  That makes them useless to catch regressions.  They'd only be a maintenance burden.

aranhoide:
@oxtoacart ^ (I don't know if you're getting email notifications for all comments here)

aranhoide:
Subsumed by multifronting-valencia.

PR #41, opened 2015-02-12
Whitelist functionality

atavism:
Still need to add extra unit tests; in particular, I want to add a mock YAML config to better test whitelist customization. I'd also like to add a test verifying sites we add to cloud.yaml end up propagating to a mock PAC file. Nevertheless, opening the PR for now to offer an overview of changes.

The original PR was 1,500 files long. I removed three packages that were showing up in the diff: src/golang.org/x/, src/code.google.com/p/freetype-go/, and the UI updates (that's on a corresponding issue-2195).

This also includes changes for https://github.com/getlantern/lantern/issues/2231 https://github.com/getlantern/lantern/issues/2205 and https://github.com/getlantern/lantern/issues/2218


src/github.com/getlantern/flashlight/config/config.go
```diff
@@ -336,5 +347,16 @@ func (updated *Config) updateFrom(updateBytes []byte) error {
...
+	// Same with global whitelist
+	// Would be nice to combine this with the above at some point!
+	wlDomains := make(map[string]bool)
+	for _, domain := range updated.Client.Whitelist.Cloud {
+		wlDomains[domain] = true
+	}
+	updated.Client.Whitelist.Cloud = make([]string, 0, len(wlDomains))
+	for domain, _ := range wlDomains {
+		updated.Client.Whitelist.Cloud = append(updated.Client.Whitelist.Cloud, domain)
+	}
```
atavism: This prevents YAML de-duplication of the global whitelist. Similar to what happens with the list of domain-fronted servers.

src/github.com/getlantern/whitelist/whitelist.go
```diff
@@ -0,0 +1,239 @@
...
+		/* need to remove escapes
+		* and convert the otto value into a string array
+		 */
+		re := regexp.MustCompile("(\\\\.)")
+		list := re.ReplaceAllString(value.String(), ".")
+		wl.entries = strings.Split(list, ",")
+		log.Debugf("List of proxied sites... %+v", wl.entries)
+	}
+	return wl
+}
```
atavism: This is mainly used to test the contents of the PAC file and could probably be moved to whitelist_test.go

myleshorton:
Hey @xiam maybe you could take a look at this?

oxtoacart:
@myleshorton @xiam I will take a look too.

atavism:
FYI, in case you notice this while testing: there's an issue using the search input on the proxied sites screen; it's probably related to stripping jQuery out. I'll let you know once a fix is deployed.

xiam:
@myleshorton Of course! Thanks @oxtoacart, this is huge!

atavism:
@xiam It has a lot of subtree code that doesn't need reviewed. I'll list the actual files that need to be looked at.

xiam:
@atavism Sure, that would help. Thanks!

This file: src/github.com/getlantern/enproxy/enproxy_linux_amd64.test is a linux executable, do we need it on the source repository? (Looks like I can't add an inline comment to it).

src/github.com/getlantern/flashlight/flashlight.go
```diff
@@ -109,6 +111,34 @@ func runClientProxy(cfg *config.Config) {
...
 	client.Configure(cfg.Client)
 
+	// this flag specifies the port to open the HTTP server
+	// between the UI and
+	// with a corresponding HTTP server at
+	// the following address
+	if cfg.Client.UiAddr != "" {
+		wlChan := make(chan *whitelist.Config)
+		go func() {
+			http.ListenAndServe(cfg.Client, configUpdates, wlChan)
```
xiam: It would be useful to catch the error that ListenAndServe returns, such as when the port is already used by another process. Otherwise we wouldn't know why the server didn't start.

If we depend on this server to continue, we may as well mark this error as fatal.

src/github.com/getlantern/lantern-android/app/src/go/Go.java
```diff
@@ -0,0 +1,47 @@
+package go;
```
xiam: The lantern-android package [hasn't been accepted yet][1]. I think we should wait until its PR gets completed before adding it to the source.

Would it be possible to `git revert` the commit that added this package instead of using `git rm`? I think it would be safer for the lantern-android's [pending PR][1].

[1]: https://github.com/getlantern/flashlight-build/pull/36

xiam:
Some other binary files that we probably don't need in the source tree:

src/github.com/getlantern/flashlight/genconfig/cloud.1.6.0.yaml.gz
src/github.com/getlantern/flashlight/genconfig/genconfig
src/github.com/getlantern/lantern-android/app/libs/armeabi-v7a/libgojni.so
 src/github.com/getlantern/peerscanner/cf/cf_linux_amd64.test.gz
src/github.com/getlantern/peerscanner/peerscanner
src/github.com/getlantern/peerscanner/peerscanner_*

src/github.com/getlantern/flashlight/genconfig/genconfig.go
```diff
@@ -98,6 +105,38 @@ func loadDomains() {
...
+	if info.IsDir() {
+		// skip root directory
+		return nil
+	}
+	whitelistBytes, err := ioutil.ReadFile(path)
+	if err != nil {
+		log.Fatalf("Unable to read blacklist file at %s: %s", path, err)
+	}
+	for _, domain := range strings.Split(string(whitelistBytes), "\n") {
+		// skip empty lines, comments, and Iranian domains
```
xiam: Why do we need to skip Iranian domains?

src/github.com/getlantern/flashlight/http/http.go
```diff
@@ -0,0 +1,164 @@
...
+		// a separate port
+		return
+	case "POST":
+		// update whitelist
+		decoder := json.NewDecoder(r.Body)
+		var entries []string
+		err := decoder.Decode(&entries)
+		if err != nil {
+			log.Error(err)
+			response.Error = fmt.Sprintf("Error decoding whitelist: %s", err.Error())
```
xiam: Just a comment, we could also use the `%q` format code to get rid of the `Error()` method on `err`:

```go
fmt.Sprintf("Error decoding whitelist: %q", err)
```

src/github.com/getlantern/flashlight/flashlight.go
xiam: I read that this is not `ListenAndServe` from the `net/http` package, but from the `github.com/getlantern/flashlight/http` and it does not return an error value, so there's nothing to catch here actually. I think it would be useful to return an error value on this situation, so we could print it and exit the application.

src/github.com/getlantern/flashlight/util/common.go
```diff
@@ -0,0 +1,62 @@
...
+
+// determineConfigDir() determines where to load the config by checking the
+// command line and defaulting to ~/.lantern.
+func DetermineConfigDir() (string, error) {
+	usr, err := user.Current()
+	if err != nil {
+		fmt.Errorf("Error location user home directory %v", err)
+		return "", err
+	}
+	return usr.HomeDir + "/.lantern", nil
```
xiam: Maybe not a top priority, but we aren't checking if `$HOME/.lantern` is a actually a directory. Also, does the `/` work on Windows too?

src/github.com/getlantern/flashlight/util/common.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+func Check(e error, log logger, msg string) {
+	if e != nil {
+		log(msg)
+	}
+}
+
+func FileExists(fileName string) (bool, error) {
```
xiam: If we need a file and not just a path, we should also call `os.FileInfo.Mode.IsRegular()`, but that would leave symlinks out, we could also check the value of the `os.FileInfo.IsDir()` method and negate it.

src/github.com/getlantern/flashlight/util/common.go
```diff
@@ -0,0 +1,62 @@
...
+
+	if !src.IsDir() {
+		err := "UI source exists, but it's not a directorY"
+		return false, errors.New(err)
+	}
+
+	return true, nil
+}
+
+func LoadTemplate(name string) string {
```
xiam: Looks like this function is defined but never used.

src/github.com/getlantern/peerscanner/mem_20150120_085312.309949401.prof
```diff
@@ -0,0 +1,601 @@
+heap profile: 16: 330528 [2713: 37329872] @ heap/1048576
```
xiam: I'm not sure if we need this file on the source tree.

src/github.com/getlantern/peerscanner/pk.pem
```diff
@@ -0,0 +1,27 @@
+-----BEGIN RSA PRIVATE KEY-----
```
xiam: We probably should not be adding a private key to a public repo, third parties could use the compromised key to forge certificates in a MITM situation. I think It would be safer to use the (fixed) certificate we're creating in `peerscanner/web.go`'s `startHttp()` and a public key (of another keypair) instead of sharing the private key and generating a certificate on the fly.

src/github.com/getlantern/tarfs/demo/resources.go
```diff
@@ -0,0 +1,3 @@
+package main
```
xiam: This file is not actual editable source code, not sure if we should add it to the source tree, we may as well generate it at build time.

src/github.com/getlantern/flashlight/flashlight.go
atavism: Agreed. I'm returning an error now and also renamed the function to avoid confusion with ```ListenAndServe``` from ```net/http```

src/github.com/getlantern/flashlight/genconfig/genconfig.go
atavism: Since we're focusing on Iran with this first release, we decided to skip adding *.ir sites to the global whitelist to avoid proxying sites that are already unblocked there. This is a general problem when you aren't maintaining country-specific whitelists. We'll probably address that in the next phase. I added a comment to make this clearer.

src/github.com/getlantern/flashlight/http/http.go
atavism: Ah, thanks! Updated.

src/github.com/getlantern/flashlight/util/common.go
atavism: Thanks! Definitely something we want to check for now. I updated that function to check for the user home directory if it doesn't already exist. I also changed that function to use path.Join() instead. If the PAC file is missing, I also create it now to synchronize it with the YAML config. Finally, I'm using https://github.com/mitchellh/go-homedir for detecting the home directory now.

src/github.com/getlantern/flashlight/util/common.go
atavism: I went ahead and removed it.

src/github.com/getlantern/peerscanner/mem_20150120_085312.309949401.prof
atavism: Thanks, definitely not. I accidentally committed a bunch of files in the peerscanner directory. I added *.prof, *.gz, and *.pem to .gitignore

src/github.com/getlantern/flashlight/client/client.go
```diff
@@ -66,7 +66,8 @@ func (client *Client) Configure(cfg *ClientConfig) {
 	log.Debug("Configure() called")
 	if client.priorCfg != nil && client.priorTrustedCAs != nil {
 		if reflect.DeepEqual(client.priorCfg, cfg) &&
-			reflect.DeepEqual(client.priorTrustedCAs, globals.TrustedCAs) {
+			reflect.DeepEqual(client.priorTrustedCAs, globals.TrustedCAs) &&
+			reflect.DeepEqual(client.priorCfg.Whitelist, cfg.Whitelist) {
```
oxtoacart: This seems redundant with `reflect.DeepEqual(client.priorCfg, cfg)` on line 68.

src/github.com/getlantern/flashlight/util/common.go
```diff
@@ -0,0 +1,61 @@
...
+// GetUserHomeDir() determines the user home directory using the go-homedir
+// package which can be used in cross-compliation environments
+func GetUserHomeDir() (string, error) {
+	homedir, err := homedir.Dir()
+	if err != nil {
+		log.Errorf("Error locating user home directory %s", err)
+		return "", err
+	}
+	lanternDir := path.Join(homedir, ".lantern")
+	// Create the user home directory if it doesn't exist already
```
oxtoacart: You mean create lanternDir instead of "user home directory"?

src/github.com/getlantern/flashlight/util/common.go
```diff
@@ -0,0 +1,61 @@
...
+	exists, _ := DirExists(lanternDir)
+	if !exists {
+		err = os.Mkdir(lanternDir, 0777)
```
oxtoacart: I don't think this should be world writeable.  0755 is probably a better mode.

src/github.com/getlantern/flashlight/util/common.go
```diff
@@ -0,0 +1,61 @@
+package util
+
+import (
+	"errors"
+	"github.com/getlantern/golog"
+	"github.com/mitchellh/go-homedir"
```
oxtoacart: Why use this instead of Go's built-in `user.Current().HomeDir` API?

src/github.com/getlantern/flashlight/util/common.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+func DirExists(dirName string) (bool, error) {
+	src, err := os.Stat(dirName)
+	if err != nil {
+		return false, err
+	}
+
+	if !src.IsDir() {
+		err := "UI source exists, but it's not a directorY"
```
oxtoacart: This function is in the common package but this error references "UI"?

src/github.com/getlantern/flashlight/client/config.go
```diff
@@ -20,6 +21,10 @@ type ClientConfig struct {
 	ChainedServers map[string]*ChainedServerInfo
 	Peers          map[string]*nattywad.ServerPeer // keyed to peer id (e.g. XMPP JID)
 	MasqueradeSets map[string][]*fronted.Masquerade
+
+	OpenUi    bool              // Flag specifying whether or not the UI should be automatically open in the browser
```
oxtoacart: Tiny nit, but can you please capitalize abbreviations, i.e. `OpenUI` ?

oxtoacart: This just controls opening the UI on start, right?

src/github.com/getlantern/flashlight/client/config.go
```diff
@@ -20,6 +21,10 @@ type ClientConfig struct {
...
+	UiAddr    string            // UI HTTP server address
```
oxtoacart: Why is the UI stuff under client?  At some point, give-mode (i.e. server) flashlight will also use the UI, right?

src/github.com/getlantern/flashlight/config/config.go
```diff
@@ -336,5 +347,16 @@ func (updated *Config) updateFrom(updateBytes []byte) error {
 	for _, server := range servers {
 		updated.Client.FrontedServers = append(updated.Client.FrontedServers, server)
 	}
+
+	// Same with global whitelist
+	// Would be nice to combine this with the above at some point!
```
oxtoacart: Might be tough to do without generics :(

src/github.com/getlantern/flashlight/client/config.go
atavism: I updated all instances of "Ui" to "UI". Yeah, only on start.

src/github.com/getlantern/flashlight/client/config.go
atavism: I just put it there for now since we're only using it in get-mode, but out of the client and into the parent config.

src/github.com/getlantern/flashlight/util/common.go
atavism: Woops, yeah. That comment should be fixed now.

src/github.com/getlantern/flashlight/genconfig/genconfig.go
```diff
@@ -69,6 +75,7 @@ func main() {
 	runtime.GOMAXPROCS(numcores)
 
 	loadDomains()
+	loadWhitelists()
```
oxtoacart: The names here could stand to be cleaned up.  Blacklist and whitelist sound like they're related, but they're not.  Blacklist is a list of domains which we don't want to use for masquerading, whereas whitelist is a list of proxied sites.  In fact, though it'll be a little bit of a pain, I'm wondering if we shouldn't refer to it as proxiedSitesList throughout the code?  Whitelist is a pretty generic term that can apply to a bunch of different things.

src/github.com/getlantern/flashlight/util/common.go
atavism: go-homedir is recommended when cross-compiling since it doesn't require the use of cgo on Darwin the way "os/user" does. Is this something we would want to use in byteexec as well?

src/github.com/getlantern/flashlight/http/http.go
```diff
@@ -0,0 +1,167 @@
...
+	"net/http"
+	"reflect"
+	"time"
+
+	"github.com/getlantern/tarfs"
+)
+
+const (
+	UiUrl = "http://%s"
+	UiDir = "src/github.com/getlantern/ui/app"
```
oxtoacart: This assumes that flashlight is run from the root of flbuild. That's a little weird since flashlight is theoretically it's own project, so if someone isn't running it in the context of flbuild, this won't work.  I tend to run it from its own source folder (i.e. src/github.com/getlantern/flashlight).  Perhaps to clear up confusion, you could add a flag `localuidir` or something like that and give it a sensible default.

src/github.com/getlantern/flashlight/genconfig/genconfig.go
atavism: Okay, that's fine. I'm not attached to the whitelist name and agree that's more specific. How about just proxiedsites instead of proxiedsiteslist for the sake of brevity?

atavism:
> FYI, in case you notice this while testing: there's an issue using the search input on the proxied sites screen; it's probably related to stripping jQuery out. I'll let you know once a fix is deployed.

This should be resolved now.

src/github.com/getlantern/flashlight/util/common.go
oxtoacart: We actually intentionally cross compile with cgo support because it improves things like DNS lookup performance.  That's why we use gonative (it has support for cross-compiling with cgo dependencies).

src/github.com/getlantern/flashlight/genconfig/genconfig.go
oxtoacart: `proxiedsites` sounds great!  Yeah, `whitelist` is historical - that's what we've been using, but this is a good opportunity to get more precise.

src/github.com/getlantern/flashlight/util/common.go
atavism: Ah, that's right.. we switched back to gonative. Okay, I went ahead and reverted that function to use user.Current().HomeDir instead.

src/github.com/getlantern/flashlight/http/http.go
```diff
@@ -0,0 +1,167 @@
...
+		}
+	case "GET":
+		response.Whitelist = wlh.whitelist.GetEntries()
+	default:
+		log.Debugf("Received %s", response.Error)
+		response.Error = "Invalid whitelist HTTP request"
+		response.Whitelist = nil
+	}
+	sendJsonResponse(w, &response, false)
+}
```
oxtoacart: If I understand this protocol correctly, the client posts the entire list every time and the back-end calculates the deltas.  This protocol will have problems if the user uses multiple windows at once.  After one window updates, the other window will have a stale proxiedsites list, and if the user makes an update there, it will overwrite whatever the user did in the other window.  However, since the other window doesn't update, there won't be anything to make that clear to the user.

A safer protocol is to have the UI calculate the deltas and post those.  The deltas are really what the user is intending.  If the user submits conflicting deltas from two different windows, that's okay because they both still represent the user's intention at the time.  A similar problem arises if someone edits the yaml file manually.

I think the best protocol would be something like this:

- Use a websocket
- Back-end posts whitelist deltas to the websocket.  On initial connection, the entire whitelist is posted as a series of adds.  For performance, we should be able to batch multiple sites in a single delta (e.g. add sites 1, 2 and 3).
- When a user makes changes to the whitelist, post those changes to the websocket as deltas (again, support batching for performance)
- Whenever the whitelist has changed, post the deltas to all connected UIs (note, this should be driven by the YAML config having been changed, not by the HTTP handler)

This will allow multiple windows and the config on disk to remain in sync.  It will also be a more useful protocol once we have an extension that allows people to add/remove the current site from the whitelist, as they can use the same API and just have a batch of 1.

oxtoacart: Actually, we can keep it simpler and use eventsource instead of websockets.  The GET request would start up the event sourcing bit, and the POST could continue to post changes, but instead of posting the whole whitelist, it could just include deltas like this:

{add: [d1.com, d2.com], remove: [d3.com, d4.com]}

@fffw Suggested an event sourcing library that I think would work well for this - https://github.com/antage/eventsource

src/github.com/getlantern/flashlight/http/http.go
```diff
@@ -0,0 +1,167 @@
...
+	log = golog.LoggerFor("http")
+)
+
+type JsonResponse struct {
+	Error     string   `json:"Error, omitempty"`
+	Whitelist []string `json:"Whitelist, omitempty"`
+	Original  []string `json:"Original, omitempty"`
+}
+
+type WhitelistHandler struct {
```
oxtoacart: I think that this should live in the whitelist library and that the existing Whitelist type should just implement the http.Handler interface.

src/github.com/getlantern/flashlight/http/http.go
```diff
@@ -0,0 +1,167 @@
...
+	http.HandlerFunc
```
oxtoacart: This is unnecessary, since you're not embedded an existing HandlerFunc.

src/github.com/getlantern/flashlight/http/http.go
atavism: Part of the reason I'm still POSTing the entire list back was to avoid having to make too many changes to the existing client-side code that handles the proxied sites configuration, but I agree changing it now would be preferable, especially from the standpoint of supporting multiple windows. In addition, given the size of the list now, it's wasteful sending everything back again when we only need to know the deltas. I'll go head and integrate eventsource.

oxtoacart: Thanks!  I think that's a good call.

atavism: One caveat is we'll already preclude support for IE if we go with SSEs. IE 10 and 11 support websockets, however.

oxtoacart: Good point.  We already use EventSource for the map, so as it stands we'll
be pretty broken in IE.

Are we supporting IE?

atavism: I thought we were. I actually have a websocket implementation working already so if there are no objections I think we should just go with that?

oxtoacart: As long as the websocket implementation handles reconnecting automatically,
I'm cool with it.

Cheers,
Ox

-----------------------------------------------------------------------------------------

"I love people who harness themselves, an ox to a heavy cart,
who pull like water buffalo, with massive patience,
who strain in the mud and the muck to move things forward,
who do what has to be done, again and again."

- Marge Piercy

src/github.com/getlantern/flashlight/util/common.go
```diff
@@ -0,0 +1,58 @@
...
+		log.Errorf("Error locating user home directory %s", err)
+		return "", err
+	}
+	lanternDir := path.Join(usr.HomeDir, ".lantern")
+	// Create the ~/.lantern directory if it doesn't exist already
+	exists, _ := DirExists(lanternDir)
+	if !exists {
+		err = os.Mkdir(lanternDir, 0755)
+		if err != nil {
+			log.Errorf("Error creating user home directory: %s", err)
```
xiam: This could be misleading, as we're creating the lantern dir and not the user dir.

src/github.com/getlantern/flashlight/util/common.go
oxtoacart: Let's not worry about it.  I introduced [appdir] in another branch (used for logging) and I'll refactor this to use it once merged.

xiam:
We're including Go dependencies in `flashlight-build`, I understand that as a safety measure against third party packages breaking their own APIs since Go packages are not versioned by default. Bower and NPM components are versioned. Since `node_components` and `bower_components` directories can be generated by npm and bower programs respectively, would we get any benefit from including (`src/github.com/getlantern/ui/app/app/bower_components`) into the source tree?

src/github.com/getlantern/flashlight/util/common.go
atavism: Thanks. I updated the error message.

src/github.com/getlantern/flashlight/http/http.go
atavism: @oxtoacart Yep, it does. I'm about finished with the websocket changes: should I create a second PR and merge into this branch? That'll actually be cleaner since there will be less files to inspect in the diff.

atavism: @oxtoacart Actually, the websocket reconnects automatically, but the UI doesn't display a message when the back-end server is suddenly unreachable. I'll make a change to display the "Lantern went away" screen again.

fffw:
@xiam I think the point is not to care about `npm install` or `bower install` at all when developing Go code.

Like @myleshorton suggested under https://github.com/getlantern/lantern-ui/pull/82, we'll take a further step using UI assets under 'dist' directly regardless of development or production.

src/github.com/getlantern/flashlight/http/http.go
atavism: Okay, I'm showing the "Waiting for Lantern" spinner again when the UI first opens and until the websocket connects. If the back-end goes away, the "Lantern appears not to be running anymore" screen appears. Finally, when/if the websocket is able to reconnect, that screen goes away.

oxtoacart:
@atavism Let me know when you'd like me to take another pass through this one.

xiam:
```
@xiam I think the point is not to care about npm install or bower install at all when developing Go code.
```

Makes sense, thanks @fffw!

Hello @atavism,

It seems like I can't leave an inline comment on this snippet from `src/github.com/getlantern/ui/app/js/services.js` so I'm leaving it here:

```
+    // but get a tracker set up and ready for use if analytics become enabled
+    // https://developers.google.com/analytics/devguides/collection/analyticsjs/field-reference
+    ga('create', GOOGLE_ANALYTICS_WEBPROP_ID, {cookieDomain: 'none'});
+    ga('set', {
+      anonymizeIp: true,
+      forceSSL: true,
+      location: 'http://lantern-ui/',
+      hostname: 'lantern-ui',
+      title: 'lantern-ui'
+    });
```

I have a bit of a doubt: I think the "lantern-ui" hostname won't work without changing user's machine configuration or DNS settings, are we already considering that in the snippet above?

And about this snippet below:

```
+    // connect() starts listening for peer updates
+    function connect() {
+      var source = new EventSource('http://127.0.0.1:15670/');
+      source.addEventListener('message', function(e) {
```

We'll probably have to deal with already used ports in some special cases, given that, declaring and using a constant instead of 'http://127.0.0.1:15670/' would probably be more useful in the future, when we deal with these special cases.

atavism:
> I have a bit of a doubt: I think the "lantern-ui" hostname won't work without changing user's machine configuration or DNS settings, are we already considering that in the snippet above?

That's the existing Google Analytics code that's just showing up in the diff. Looking at the stats, we don't really factor in hostname or location so I think we can probably safely remove those fields actually. 

> We'll probably have to deal with already used ports in some special cases, given that, declaring and using a constant instead of 'http://127.0.0.1:15670/' would probably be more useful in the future, when we deal with these special cases.

That's the client stats address that flashlight uses, and it isn't expected to change. That snippet isn't active at the moment either since we haven't implemented stats yet. I think we can just leave that in-place for now.

oxtoacart:
> That's the client stats address that flashlight uses, and it isn't expected to change. That snippet isn't active at the moment either since we haven't implemented stats yet. I think we can just leave that in-place for now.

Actually, feeding stats to the map is already implemented.  See https://github.com/getlantern/flashlight-build/blob/devel/src/github.com/getlantern/flashlight/statserver/statserver.go and https://github.com/getlantern/lantern-ui/blob/master/app/js/services.js#L286.

atavism:
> Actually, feeding stats to the map is already implemented. See https://github.com/getlantern/flashlight-build/blob/devel/src/github.com/getlantern/flashlight/statserver/statserver.go and https://github.com/getlantern/lantern-ui/blob/master/app/js/services.js#L286.

But "127.0.0.1:15670" is hard-coded. We need to send the stats address to the UI still when you run flashlight, right?

atavism:
Also, if we're using websockets now and want to support IE, we probably need to update that code to not use eventsource anyway.

oxtoacart:
> But "127.0.0.1:15670" is hard-coded. We need to send the stats address to the UI still when you run flashlight, right?

Not sure what you mean.  In Java Lantern, we tell flashlight to expose stats on port 15670 and the UI is hardcoded to talk to that port for stats.

> Also, if we're using websockets now and plan to support IE, we need to update that code anyway.

@myleshorton Is IE support a requirement for the Valencia release?  If not, we can leave it as is for now.

atavism:
> Not sure what you mean. In Java Lantern, we tell flashlight to expose stats on port 15670 and the UI is hardcoded to talk to that port for stats.

I can run flashlight and pass in a different statsaddr: the UI won't know what port the stats server is running on. Since port "15670" is hard-coded, the UI will expect events to arrive on a different address.

oxtoacart:
> I can run flashlight and pass in a different statsaddr: the UI won't know what port the stats server is running on. Since port "15670" is hard-coded, the UI will expect events to arrive on a different address.

Gotcha. Once your UI changes are ready, I'll update the statserver to expose its API on the same HTTP server as the whitelist, just under a different path.  Then we can just get rid of the statsaddr setting.

oxtoacart:
@atavism I'm going to merge the websocket branch directly into valencia in lieu of this.  I hope that's alright.

PR #42, opened 2015-02-14
Closes getlantern/lantern#2216 Tracking and managing orange cloud status...

oxtoacart:
... separately from dns record

### DIFF
--- a/src/github.com/getlantern/peerscanner/cf/cf.go
+++ b/src/github.com/getlantern/peerscanner/cf/cf.go
@@ -70,7 +70,7 @@ func (util *Util) EnsureRegistered(name string, ip string, rec *cloudflare.Recor
 		rec, err = util.Client.CreateRecord(util.domain, &cr)
 
 		if err != nil {
-			if !isDuplicateError(err) {
+			if !isDuplicateRecord(err) {
 				return nil, false, err
 			}
 			log.Debugf("%v (%v) already registered, looking up existing record", name, ip)
@@ -114,6 +114,6 @@ func (util *Util) DestroyRecord(r *cloudflare.Record) error {
 	return util.Client.DestroyRecord(util.domain, r.Id)
 }
 
-func isDuplicateError(err error) bool {
+func isDuplicateRecord(err error) bool {
 	return strings.Contains(err.Error(), "The record already exists.")
 }
--- a/src/github.com/getlantern/peerscanner/cf/cf.go
+++ b/src/github.com/getlantern/peerscanner/cf/cf.go
@@ -5,6 +5,7 @@ import (
 	"crypto/tls"
 	"fmt"
 	"net/http"
+	"strings"
 	"time"
 
 	"github.com/getlantern/cloudflare"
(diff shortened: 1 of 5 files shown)


oxtoacart:
I've already deployed a preview of this to production and am watching the logs and testing with a client.

src/github.com/getlantern/peerscanner/cf/cf.go
```diff
@@ -53,31 +54,66 @@ func (util *Util) GetAllRecords() ([]cloudflare.Record, error) {
...
+	return rec, true, nil
 }
 
 func (util *Util) DestroyRecord(r *cloudflare.Record) error {
 	return util.Client.DestroyRecord(util.domain, r.Id)
 }
+
+func isDuplicateError(err error) bool {
+	return strings.Contains(err.Error(), "The record already exists.")
+}
```
myleshorton, line 119: This name confused me -- made me think it was looking for duplicate _errors_ instead of duplicate _entries_. Maybe just isDuplicateEntry?

src/github.com/getlantern/peerscanner/cf/cf.go
```diff
@@ -53,31 +54,66 @@ func (util *Util) GetAllRecords() ([]cloudflare.Record, error) {
...
+		if err != nil {
+			if !isDuplicateError(err) {
+				return nil, false, err
+			}
+			log.Debugf("%v (%v) already registered, looking up existing record", name, ip)
+			// Note - this is pretty heavyweight since it fetches all
+			// records, but this condition should rarely be hit anyway
+			all, err := util.GetAllRecords()
+			if err != nil {
+				return nil, false, err
```
myleshorton, line 81: Maybe log this? Do we have any way of seeing error logs on this guy in general? How easy would it be to link it up with Papertrail if we don't currently have a way to see 'em?

src/github.com/getlantern/peerscanner/cf/cf.go
```diff
@@ -53,31 +54,66 @@ func (util *Util) GetAllRecords() ([]cloudflare.Record, error) {
...
+		// Register record
+		var err error
+		cr := cloudflare.CreateRecord{Type: "A", Name: name, Content: ip}
+		rec, err = util.Client.CreateRecord(util.domain, &cr)
 
-	if err != nil {
-		return nil, err
+		if err != nil {
+			if !isDuplicateError(err) {
+				return nil, false, err
```
myleshorton, line 74: Maybe also log it if it's not a duplicate entry? Would be useful to know if there's some other error happening.

myleshorton:
Way to knock this out @oxtoacart! I just commented on a couple of things but otherwise looks good. I'm mostly concerned with our ability to catch and notice errors in production if they occur.

src/github.com/getlantern/peerscanner/cf/cf.go
oxtoacart, line 74: The error already gets logged higher up where it's handled.

src/github.com/getlantern/peerscanner/cf/cf.go
oxtoacart, line 81: The error already gets logged higher up.

src/github.com/getlantern/peerscanner/cf/cf.go
oxtoacart, line 119: Fixed.

src/github.com/getlantern/peerscanner/cf/cf.go
myleshorton, line 81: To papertrail is that right @oxtoacart?

oxtoacart, line 81: No, this is all just going to syslog on that machine right now.  After ssh'ing to the server:

`sudo grep peerscanner /var/log/syslog`

myleshorton, line 81: OK cool. I think ultimately we should figure out a common way for all projects like this to log probably to Loggly so we can consistently catch errors, but looks fine to me otherwise.

PR #43, opened 2015-02-15
Update README.md

fffw:
@oxtoacart  we should switch back ,right?
Also changed go version to 1.4.1


oxtoacart:
I already fixed this here - https://github.com/getlantern/flashlight-build/blob/valencia-mac/README.md

PR #47, opened 2015-02-16
Add logging to files with rotation

oxtoacart:


src/github.com/dogenzaka/rotator/.gitignore
```diff
@@ -0,0 +1,24 @@
+# Compiled Object files, Static and Dynamic libs (Shared Objects)
```
oxtoacart: rotator is a 3rd party library that we use for log rotation

src/github.com/getlantern/flashlight/flashlight.go
```diff
@@ -169,6 +177,35 @@ func useAllCores() {
 	runtime.GOMAXPROCS(numcores)
 }
 
+func configureLogging() *rotator.SizeRotator {
+	logdir := appdir.Logs("Lantern")
+	log.Debugf("Placing logs in %v", logdir)
+	if _, err := os.Stat(logdir); err != nil {
+		if os.IsNotExist(err) {
```
fffw: I don't know how, but should also check if directory is writable. Just create a file to test?
> **drwxr-xr-x**  5 root  staff  170 Jan 15 19:47 /Users/merlin/Library/Logs/Lantern/

src/github.com/getlantern/flashlight/flashlight.go
```diff
@@ -169,6 +177,35 @@ func useAllCores() {
...
+				log.Fatalf("Unable to create logdir at %s: %s", logdir, err)
+			}
+		}
+	}
+	file := rotator.NewSizeRotator(filepath.Join(logdir, "lantern.log"))
+	// Set log files to 1 MB
+	file.RotationSize = 1 * 1024 * 1024
+	// Keep up to 20 log files
+	file.MaxRotation = 20
+	errorOut := timestamped(io.MultiWriter(file, os.Stderr))
```
fffw: MultiWriter will stop writing on first error, we'd better to write to Stderr first.

src/github.com/getlantern/flashlight/flashlight.go
```diff
@@ -169,6 +177,35 @@ func useAllCores() {
...
+	debugOut := timestamped(io.MultiWriter(file, os.Stdout))
```
fffw: Does it make sense to use separate rotator for errorOut and debugOut? Maybe lantern.err and lantern.log.

src/github.com/getlantern/flashlight/flashlight.go
```diff
@@ -169,6 +177,35 @@ func useAllCores() {
...
+	golog.SetOutputs(errorOut, debugOut)
+	return file
+}
+
+// timestamped adds a timestamp to the beginning of log lines
+func timestamped(orig io.Writer) io.Writer {
+	return wfilter.Lines(orig, func(w io.Writer, line string) (int, error) {
+		return fmt.Fprintf(w, "%s - %v", time.Now().Format(time.RFC3339), line)
```
fffw: I prefer change formater to `time.StampMilli`, so we get a more compact and accurate timestamp.
> Feb 17 07:29:57.160 - ...

fffw: And I even prefer to add this timestamp into golog, as it's a must for any logging facility.

src/github.com/getlantern/flashlight/flashlight.go
fffw: >  -rw-r--r--  1 root    staff  2032 Jan 15 19:47 installer.log
>  -rw-r--r--  1 root    staff     0 Jan 15 19:47 install4jout.txt
>  -rw-r--r--  1 root    staff     0 Jan 15 19:47 install4jerr.txt

It was created during installation of Lantern java version, so it's vital to ensure writability. We may need to fix it in installer, or just remove and recreate whole directory if not writable.

oxtoacart: No need.  If there's a problem writing, there'll be an error on write.  This was getting swallowed, but I fixed golog to log to stderr when it happens.

Here's an example - `Unable to log: open /usr/bin/lantern: permission denied`

Regarding the legacy Lantern version, we said on our Friday call that we would not worry about the use case of someone who had the old Lantern installing this new Lantern.  If we change our mind, this and a whole bunch of other things will need consideration.

src/github.com/getlantern/flashlight/flashlight.go
oxtoacart: Good catch!  Fixed.

src/github.com/getlantern/flashlight/flashlight.go
oxtoacart: No.  By writing to the same file, we can see errors in the context of corresponding debug log information.

src/github.com/getlantern/flashlight/flashlight.go
oxtoacart: I chose the formatter that I did because it is constant width and preserves alignment of the log messages, which helps readability.

I do like the more compact representation, so I am using a fixed-width version of it:

`Jan 02 15:04:05.000`

I'm also converting the time to UTC before logging it so that we don't have to bother including a timezone.  This is important because we'll be getting log messages from all over the world.

oxtoacart: I won't this functionality to golog because I'm following the Go philosophy of keeping libraries small and focused.  golog is just an API for capturing log information at different log levels and writing it to somewhere.  As the `configureLogging` method shows, it is trivial to add behavior like timestamp formatting, log rotation and eventually Loggly by composing writers.

src/github.com/getlantern/wfilter/wfilter.go
```diff
@@ -0,0 +1,66 @@
...
+	return &lp{w, prepend, true}
+}
+
+type lp struct {
+	io.Writer
+	prepend       Prepend
+	prependNeeded bool
+}
+
+func (w *lp) Write(buf []byte) (int, error) {
```
fffw: I was trying very hard to understand this method but failed... So I write a shorter `do... while()` version, please take a look to check if any flaw exists.

fffw:
Understood the code finally, very cool, merging.

oxtoacart:
@fffw Thanks for reviewing!

PR #49, opened 2015-02-16
Tracking down some issues with non-fallbacks getting registered as fallb...

oxtoacart:
...acks

### DIFF
--- a/src/github.com/getlantern/peerscanner/host.go
+++ b/src/github.com/getlantern/peerscanner/host.go
@@ -262,16 +262,20 @@ func (h *host) reportStatus(s *status) {
 
 func (h *host) doReset(newName string) {
 	log.Tracef("Host notified us of its presence")
-	h.lastSuccess = time.Now()
-	h.lastTest = time.Time{}
 	if newName != h.name {
 		log.Debugf("Hostname for %v changed to %v", h, newName)
 		if h.record != nil {
 			log.Debugf("Deregistering old hostname %v", h.name)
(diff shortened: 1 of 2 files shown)


oxtoacart:
@myleshorton Thanks for reviewing the PR.  I found another issue that was causing problems, namely peers sometimes stealing the address of a fallback.  I think perhaps the peers are proxying through the fallback when reporting themselves to peerscanner.  I've added some logic to protect against that.

I also found a small issue in the logic for changing a hosts name, specifically that we ignore when the deregister fails.  Really, we need to abort the name change if the deregister fails.  Assuming the proxy is still up, it will phone home again eventually at which point we'll try the name change again.

myleshorton:
I.e. Even if they're tunneling to peerscanner, peerscanner should still use the reported ip not the ip of the incoming connection. Has that changed?

oxtoacart:
> I.e. Even if they're tunneling to peerscanner, peerscanner should still use the reported ip not the ip of the incoming connection. Has that changed?

Yes, that changed.

oxtoacart:
> Yes, that changed.

The reasoning, btw, is to prevent malicious requestors from messing with proxies that aren't theirs.  Accepting X-Forwarded-For of course allows them to do that anyway, but new clients are actually contacting peerscanner.getiantem.org directly, and the goal is to phase out intervening proxies so that we can eventually rely just on the IP.

myleshorton:
OK thanks @oxtoacart. Code change looks fine.

PR #51, opened 2015-02-17
Pull changes to support building OS X App Bundle and DMG

oxtoacart:
@xiam Do you mind taking a look at this since it's related to the auto-update work you're doing?


oxtoacart:
Oops, wrong target branch.

PR #55, opened 2015-02-17
Feature/autoupdate

oxtoacart:
@xiam I'm opening this so that I can review auto-updates.


src/github.com/getlantern/autoupdate/_test_app/README.md
```diff
@@ -0,0 +1,128 @@
...
+greatest number than `n` is always defined by `n + 1` and we'll have no chance
+of hitting a bug derived from comparison of floating point values.
+
+```sh
+equinox release --config equinox.yaml --version=1 main.go
+# ...
+```
+
+Note: I've found my `equinox` tool does not actually cares about the
+`equinox.yaml` file. So I feed these values directly:
```
oxtoacart: I ran into the same issue.  Maybe you can just remove the example equinox.yaml.

src/github.com/getlantern/autoupdate/autoupdate.go
```diff
@@ -0,0 +1,169 @@
...
+		AppId:      a.cfg.appID,
+		// Should we pick an update channel from ENV? It could be useful to test
+		// development updates.
+		Channel: a.cfg.updateChannel,
+	}
+
+	up = update.New()
+
+	// TODO: This is not working.
+	// up = update.New().ApplyPatch(update.PATCHTYPE_BSDIFF)
```
xiam: Looks like I'm doing something wrong here, the `PATCHTYPE_BSDIFF` fails when `update` is configured to use signatures. I'll still need to go through the `update` package to see exactly what's happening.

src/github.com/getlantern/autoupdate/_test_app/README.md
xiam: I created [a tool](https://github.com/getlantern/autoupdate/tree/master/releasetool) to replace the equinox cli, I think it would be better if I remove the references to the old equinox cli and document usage of our `releasetool`.

xiam:
@oxtoacart Before you continue, please let me update the README, that was the first version and some improvements (such as the usage of `releasetool` and examples with signatures) are still undocumented.

oxtoacart:
@xiam 

> Before you continue, please let me update the README, that was the first version and some improvements (such as the usage of releasetool and examples with signatures) are still undocumented.

Okay.

src/github.com/getlantern/autoupdate/autoupdate.go
```diff
@@ -0,0 +1,169 @@
...
+	v   int
+	// When a patch has been applied, the patch's version will be sent to
+	// UpdatedTo.
+	UpdatedTo chan int
+}
+
+// New creates an AutoUpdate struct based in the configuration defined in
+// config.go.
+func New(appName string) *AutoUpdate {
+	if configMap[appName] == nil {
```
oxtoacart: Why have a configMap instead of just having the config passed in to this function?  Using the configMap like this requires me to actually change the source code of autoupdate in order to add my application, right?

src/github.com/getlantern/autoupdate/releasetool/client/client.go
```diff
@@ -0,0 +1,144 @@
+package client
```
xiam: This is a release tool for us, that why it does not use a proxy.

src/github.com/getlantern/autoupdate/releasetool/main.go
```diff
@@ -0,0 +1,165 @@
...
+	flag.Parse()
+
+	// Validating version.
+	if *flagVersion < 0 {
+		log.Fatal("Version must be a positive integer (-version).")
+	}
+
+	// Validating arch.
+	switch *flagArch {
+	case "amd64", "386":
```
xiam: Architectures we're supporting.

src/github.com/getlantern/autoupdate/releasetool/main.go
```diff
@@ -0,0 +1,165 @@
...
+		// OK.
+	default:
+		log.Fatal("Missing a valid build architecture (-arch).")
+	}
+
+	// Validating OS.
+	switch *flagOS {
+	case "linux", "darwin", "windows":
```
xiam: Operating systems we're supporting.

src/github.com/getlantern/autoupdate/autoupdate.go
```diff
@@ -0,0 +1,169 @@
...
+	sleepTime = time.Hour * 4
+)
+
+// SetProxy sets the proxy to use.
+func SetProxy(proxyAddr string) {
+	var err error
+
+	if proxyAddr != "" {
+		// Create a HTTP proxy and pass it to the update package.
+		if update.HTTPClient, err = util.HTTPClient("", proxyAddr); err != nil {
```
xiam: We're forcing the update package to use a proxy to communicate with equinox.

src/github.com/getlantern/autoupdate/autoupdate.go
xiam: Thats right. The original reasoning behind having the autoupdates configuration within the `autoupdate` package and outside the application that is going to use it, is to reduce the amount of code that depends on the structure of `autoupdate` given that it could change, for example, if we ever decide changing our backend and not using Equinox.IO that configuration struct would surely change.

I think this is the best way to keep packages completely independent from Equinox.IO, since the configuration struct is modelled after the Equinox.IO platform's needs. In fact, packages that import `autoupdate` do not need to know that they're going to communicate with Equinox.IO at all.

At this time we're targeting only one application, the Lantern binary, so changing the configuration struct would not be that hard if we decide to move the configuration outside the `autoupdate` package.

oxtoacart: Yeah, I think the need to decouple from equinox at the API level isn't really that important, since there are steps that the user has to take outside of using autoupdate in order to even publish updates, which are all very equinox specific.  I would just accept the configMap data structure as is for now.

xiam:
I found the reason for the `update.PATCHTYPE_BSDIFF` failing, the reason was to do with my testing and not with the go-update package itself. Basically I was asking the server to generate a binary diff against a release version that was not uploaded to the server, and that could never work of course.

Another thing that we should consider, currently the Equinox.IO service that computes the binary diffs is not working:

```
{"Error":"Failed to create patch: fork/exec /bin/bzip2: cannot allocate memory"}
```

In case of errors like the above one, go-update keeps going and just downloads the full version, so the program gets updated anyway. It's just something for us to consider.

I also added the missing documentation for `releasetool`, so this PR is ready for review.

oxtoacart:
> I found the reason for the update.PATCHTYPE_BSDIFF failing, the reason was to do with my testing and not with the go-update package itself. Basically I was asking the server to generate a binary diff against a release version that was not uploaded to the server, and that could never work of course.

Excellent!

> Another thing that we should consider, currently the Equinox.IO service that computes the binary diffs is not working:

> {"Error":"Failed to create patch: fork/exec /bin/bzip2: cannot allocate memory"}

> In case of errors like the above one, go-update keeps going and just downloads the full version, so the program gets updated anyway. It's just something for us to consider.

Now that you've had a chance to work with it a little bit, what's your gut feeling about equinox.io?  Are we taking a big risk by using it?  What would you recommend?

> I also added the missing documentation for releasetool, so this PR is ready for review.

I'm looking forward to reviewing it!

src/github.com/getlantern/autoupdate/autoupdate.go
xiam: Maybe this is not useful anymore, with the recent drop of equinox we don't need an API key or an user ID, just a public key. Do you think it would be better to embed it into the application that is going to use autoupdates?

xiam:
My latest set of changes removed the dependency we had on the equinox.io server!

src/github.com/getlantern/autoupdate/config.go
```diff
@@ -0,0 +1,15 @@
...
+
+const (
+	updateURI = `http://127.0.0.1:9197/update`
+)
+
+type config struct {
+	publicKey []byte
+}
+
+var configMap = map[string]*config{
```
xiam: Maybe this could be integrated into the application, as we have removed the client and application IDs.

src/github.com/getlantern/autoupdate-server/server/signature.go
```diff
@@ -0,0 +1,85 @@
+package server
+
+import (
+	"crypto"
+	"crypto/rand"
+	"crypto/rsa"
+	"crypto/x509"
+	"encoding/hex"
+	"encoding/pem"
```
oxtoacart: In case it's of any use to you, you can use our keyman library for doing crypto/certificate stuff - https://github.com/getlantern/keyman

src/github.com/getlantern/autoupdate/config.go
oxtoacart: Yes please!  I'd actually say this is a prerequisite before merging.

src/github.com/getlantern/autoupdate-server/server/server.go
```diff
@@ -0,0 +1,153 @@
...
+	// No update available.
+	if VersionCompare(p.AppVersion, update.v) != Higher {
+		return nil, ErrNoUpdateAvailable
+	}
+
+	// A newer version is available!
+
+	// Generate a binary diff of the two assets.
+	var patch *Patch
+	if patch, err = GeneratePatch(current.URL, update.URL); err != nil {
```
oxtoacart: It looks like this is getting called for every request.  That seems like it would get expensive.

oxtoacart:
I'm getting a test failure:

```
--- FAIL: TestSplitUpdateAsset (0.00s)
	github_test.go:16: Failed to get asset info: "Could not find asset info."
```

src/github.com/getlantern/autoupdate-server/server/bindiff.go
```diff
@@ -0,0 +1,128 @@
...
+	if err := cmd.Run(); err != nil {
+		return "", fmt.Errorf("Failed to generate patch with bsdiff: %q", err)
+	}
+
+	return patchfile, nil
+}
+
+// GeneratePatch compares the contents of two URLs and generates a patch.
+func GeneratePatch(oldfileURL string, newfileURL string) (p *Patch, err error) {
+	p = new(Patch)
```
oxtoacart: This has a race condition, since multiple calls with the same parameters will be writing to the same file.

src/github.com/getlantern/autoupdate-server/server/signature.go
xiam: Thanks! it seems like I could borrow the `LoadPKFromFile` from keyman but I'll have no use for most of the other functions. I currently use just two crypto-related functions (reading key and signing a checksum) so maybe it's not that necessary at this time.

src/github.com/getlantern/autoupdate-server/server/server.go
xiam: Great catch! Let me think on it a bit.

src/github.com/getlantern/autoupdate-server/server/bindiff.go
xiam: Got it, working on it.

xiam:
Thanks for helping me review this @oxtoacart! tests on the server package were failing. I also pushed some changes to move configuration outside the autoupdate package.

oxtoacart:
@xiam So I think the infrastructure piece of this is good.  Do you want to go ahead and integrate it into lantern before we merge?

xiam:
@oxtoacart Sure! I'll prepare a test environment for checking out our integration with flashlight.

oxtoacart:
@xiam It just occurred to me that it would be nice to compress the diffs.  You can use http://golang.org/pkg/compress/bzip2/ for that.

xiam:
@oxtoacart I also think it could very useful to compress communications and files, I already enabled gzip in the front webserver:

```
curl  -H "Accept-Encoding: gzip"  https://update.getlantern.org/patches/2ee1324ee1ae42754afae17a8c6b7aeffdf72d4d771edd27443a1626f6faa820 --head
HTTP/1.1 200 OK
Server: nginx
Date: Tue, 24 Mar 2015 20:07:38 GMT
Content-Type: application/octet-stream
Connection: keep-alive
Last-Modified: Tue, 24 Mar 2015 19:40:37 GMT
Content-Encoding: gzip
```

So maybe there's no need to implement bzip2 compression within our application, as the message is already being compressed at the HTTP level.

oxtoacart:
@xiam Yeah, if you're already compressing at the HTTP level there's no need to compress the patch itself.

Is your next step to test with the full Lantern.app/installer?

xiam:
@oxtoacart Yeah! that would be next.

myleshorton:
OK, @xiam, I just did a bunch of stuff on the scripting side to automate this with each release. Binaries currently in flashlight-build but can change.

xiam:
@myleshorton Thanks, sounds good. I'm looking into it.

xiam:
OSX: ✓

I created the DMG installer with the (signed) binary, then it ran it and installed Lantern, this is my initial state:

```
cd /Applications/Lantern.app/Contents/MacOS

shasum -a 256 lantern
bd11d24ea599ea1bc73a3d9cd98473f743beb14a9da9dcf132fee4740a22d826  lantern

codesign -dv lantern
Executable=/Applications/Lantern.app/Contents/MacOS/lantern
Identifier=com.getlantern.lantern
Format=bundle with Mach-O thin (x86_64)
CodeDirectory v=20200 size=57626 flags=0x0(none) hashes=2874+3 location=embedded
Signature size=8548
Timestamp=Mar 25, 2015, 1:57:01 PM
Info.plist entries=11
TeamIdentifier=ACZRKC3LQ9
Sealed Resources version=2 rules=12 files=1
Internal requirements count=2 size=568
```

The numbers at the bottom of the map were: `195626`

Few seconds after starting up the binary got changed:

```
shasum -a 256 lantern
8c0d9bfbf50c4d6e34565d4a49372ec831c549a8b7a7ac70891850d9727614e9  lantern
```

Permissions were kept:

```
ls -l lantern
-rwxr-xr-x  1 reventlov  admin  11836128 Mar 25 14:02 lantern
```

And codesign reported a different timestamp:

```
codesign -dv lantern
Executable=/Applications/Lantern.app/Contents/MacOS/lantern
Identifier=com.getlantern.lantern
Format=bundle with Mach-O thin (x86_64)
CodeDirectory v=20200 size=57626 flags=0x0(none) hashes=2874+3 location=embedded
Signature size=8548
Timestamp=Mar 25, 2015, 1:50:42 PM
Info.plist entries=11
TeamIdentifier=ACZRKC3LQ9
Sealed Resources version=2 rules=12 files=1
Internal requirements count=2 size=568
```

I closed lantern and opened it again, the numbers changed to:

195014

I uploaded a signed DMG release to here: https://github.com/getlantern/autoupdate/releases/tag/2.0.0-beta2 which you can use to do the same test I did. Actually, can you help me checking if the same thing works for you?

About how to apply the update:

At this stage, a natural restart initiated by the user sounds to me like the most transparent way to upgrade to a newer version, it does not require closing lantern and opening it again, nor it does require bothering the user with a message, it just works with no need of user intervention: the next time the user restarts its computer or starts Lantern the most recent version of lantern will be used.

oxtoacart:
Sweet!  The fact that Lantern.app isn't owned by root:wheel may actually be
a bug with how my scripts are building the DMG installer, as the intention
was for it to be owned by root:wheel.  That said, if we're not having any
problems because of it not being owned by root:wheel, maybe it's okay to
leave it as is.

I think that for the current Get-mode only Lantern, it's probably good
enough for the update to simply take effect whenever the user next starts
Lantern.  However for Give-mode, Lantern is really meant as a run it and
forget it kind of thing, so we'll need to figure out something else there.

Cheers,
Ox

-----------------------------------------------------------------------------------------

"I love people who harness themselves, an ox to a heavy cart,
who pull like water buffalo, with massive patience,
who strain in the mud and the muck to move things forward,
who do what has to be done, again and again."

- Marge Piercy

myleshorton:
Yeah that sequence appears to have worked for me as well, @xiam, although I got a different hash on my binary after the update. Did you modify the update between then and now? I think the permissions look fine and maybe preferable, as they match up with Chrome:

[afisk@Adams-MacBook-Pro-3 MacOS]$ ls -la /Applications/Lantern.app
total 0
drwxr-xr-x@  3 afisk  admin   102 Mar 25 12:57 .
drwxrwxr-x+ 68 root   admin  2312 Mar 25 13:27 ..
drwxr-xr-x@  7 afisk  admin   238 Mar 25 12:57 Contents
afisk@Adams-MacBook-Pro-3 MacOS]$ ls -la /Applications/Google\ Chrome.app
total 0
drwxrwxr-x@  3 afisk  admin   102 Mar 20 14:18 .
drwxrwxr-x+ 68 root   admin  2312 Mar 25 13:27 ..
drwxrwxr-x@  8 afisk  admin   272 Mar 20 14:35 Contents
[afisk@Adams-MacBook-Pro-3 MacOS]$ ls -la /Applications/Lantern.app/Contents/MacOS/lantern
-rwxr-xr-x  1 afisk  admin  11836128 Mar 25 13:31 /Applications/Lantern.app/Contents/MacOS/lantern
[afisk@Adams-MacBook-Pro-3 MacOS]$ ls -la /Applications/Google\ Chrome.app/Contents/MacOS/Google\ Chrome
-rwxrwxr-x  1 afisk  admin  17968 Mar 20 14:36 /Applications/Google Chrome.app/Contents/MacOS/Google Chrome

xiam:
@myleshorton 

Oh! I think I forgot to update the latest hash, do you have this one?:

```
3b68463eb127f0489dd7c36d335f09c657bb85a78b36b138e9d108462ea69d13  lantern
```

@oxtoacart, @myleshorton,

On the permissions issue, does that permission configuration degrade your experience with Lantern? My installation works fine. I'd agree on leaving it as is if it does not produce any weird behaviour.

myleshorton:
That's the one!

```
[afisk@Adams-MacBook-Pro-3 MacOS]$ shasum -a 256 lantern
3b68463eb127f0489dd7c36d335f09c657bb85a78b36b138e9d108462ea69d13  lantern
```

I think the permissions look good, and I agree that for now we should just wait for the user to restart.

xiam:
Windows XP: ✓

I created and signed a [Lantern installer for Windows](https://github.com/getlantern/autoupdate/releases/download/2.0.0-beta2/lantern-installer.exe), along with the auto update binary, see the full list of mock assets for this fake release here: https://github.com/getlantern/autoupdate/releases/tag/2.0.0-beta2

The lantern.exe binaries have this MD5 checksums:

```
md5 autoupdate-binary-windows-386-b*
MD5 (autoupdate-binary-windows-386-b2.exe) = 8cace5d73112e55e3ea48f0e9a5a1bbb
MD5 (autoupdate-binary-windows-386-b3.exe) = faf509bfe1da4be94043e99eec378706
```

Where the `b2` suffix means that it corresponds to the beta 2, and `b3` means beta 3.

So, I installed our Lantern beta2 via the installer, and verified the MD5 checksum with the [FCIV](http://support.microsoft.com/en-us/kb/841290) program, then I started Lantern and waited half a minute before stopping it, then checked again the MD5 sums and ran Lantern again with no problems.

The MD5 sums changed to match the expected ones:

screen shot 2015-03-26 at 9 37 21 am

I will run the same test on Windows 7.

teamlantern:
Very cool! So it's not having any problems writing to program files huh?

On Thu, Mar 26, 2015 at 8:47 AM, Carlos Nieto <notifications@github.com>
wrote:

> Windows XP: ✓
>
> I created and signed a Lantern installer for Windows
> <https://github.com/getlantern/autoupdate/releases/download/2.0.0-beta2/lantern-installer.exe>,
> along with the auto update binary, see the full list of mock assets for
> this fake release here:
> https://github.com/getlantern/autoupdate/releases/tag/2.0.0-beta2
>
> The lantern.exe binaries have this MD5 checksums:
>
> md5 autoupdate-binary-windows-386-b*
> MD5 (autoupdate-binary-windows-386-b2.exe) = 8cace5d73112e55e3ea48f0e9a5a1bbb
> MD5 (autoupdate-binary-windows-386-b3.exe) = faf509bfe1da4be94043e99eec378706
>
> Where the b2 suffix means that it corresponds to the beta 2, and b3 means
> beta 3.
>
> So, I installed our Lantern beta2 via the installer, and verified the MD5
> checksum with the FCIV <http://support.microsoft.com/en-us/kb/841290>
> program, then I started Lantern and waited half a minute before stopping
> it, then checked again the MD5 sums and ran Lantern again with no problems.
>
> The MD5 sums changed to match the expected ones:
>
> [image: screen shot 2015-03-26 at 9 37 21 am]
> <https://cloud.githubusercontent.com/assets/385670/6850101/c310b3e0-d39c-11e4-81eb-ffddba27bda3.png>
>
> I will run the same test on Windows 7.
>
> —
> <https://github.com/getlantern/flashlight-build/pull/55#issuecomment-86583880>
> .
>



-- 
--
President
https://www.getlantern.org

oxtoacart:
Windows XP is easy because your user is likely an Admin and it doesn't have UAC :)  I'm curious to see how this works on Windows 8.

teamlantern:
Oh right good point @oxtoacart!

On Thu, Mar 26, 2015 at 9:13 AM, oxtoacart <notifications@github.com> wrote:

> Windows XP is easy because your user is likely an Admin and it doesn't
> have UAC :) I'm curious to see how this works on Windows 8.
>
> —
> <https://github.com/getlantern/flashlight-build/pull/55#issuecomment-86598745>
> .
>



-- 
--
President
https://www.getlantern.org

xiam:
Yeah @oxtoacart is right, XP does not care about security permissions.

Some sad news: this does not work on Windows 7 and I doubt it works on 8 and such.

I'm having a bit of a problem here, for some reason my system becomes a bit unstable when running a Windows 7 vm and the emulation performs dog slow. I have an old netbook at home that I could use to run Windows 7 on real hardware but I'm afraid that would take a bit more time as I have to backup that netbook first. I think that I can continue checking what happens with Lantern Installer on Linux to use the time while I clean that netbook.

myleshorton:
Sounds good @xiam. We've had issues with VirtualBox VMs in the past as well and have found vmware to be much better. If you want to purchase vmware and then work with @carsonefv to get reimbursed, that could also theoretically solve it.

myleshorton:
I still think we'll ultimately have to install the executable to AppData for this to work or to set the permissions on the executable in Program Files to the current user during install.

xiam:
@myleshorton I would swear I left a comment regarding your suggestion on AppData here but now I can't find it, I was basically saying that we should look for the easiest way to fix this permission problem and your suggestion sounds pretty accurate, so I'll try that first.

I just checked this on Ubuntu Linux too, it didn't work, I was already expecting that as the binary is located at /usr/bin and requires root permissions for writing. I think we can move the lantern binary to `~/bin/` and set it with the current user's ownership. I'm not sure if the `.deb` packaging allows installing stuff for users in their ~/bin directory... maybe I should investigate how Chrome does this.

screen shot 2015-03-26 at 12 34 35 pm

Ultimately, we have the option of using the elevate package, that may work but I'd prefer not to rely that much on administrator privileges as it may pave the road to future security issues.

xiam:
Linux: ✓

Few changes on my last push:

* `/usr/bin/lantern` is now a symlink of `/usr/lib/lantern/lantern.sh`
* `/usr/lib/lantern/lantern.sh` is a wrapper of the `~/.lantern/bin/lantern` executable, if `~/.lantern/bin/lantern` does not exists, `/usr/lib/lantern/lantern.sh` copies it from `/usr/lib/lantern/lantern-binary`

The fact that `~/.lantern/bin/lantern` is copied from `/usr/lib/lantern/lantern-binary` allows the `~/.lantern/bin/lantern` to be writable by the current user, which is what we need in order for the auto-updates to work.

This is the sum of the original version bundled with the mock 2.0.0-beta2 installer:

```
sha1sum /usr/lib/lantern/lantern-binary
95a7c502bf139eafaa06a27065affd2a1e05c5ee  /usr/lib/lantern/lantern-binary
```

After a few seconds, the binary got updated:

```
sha1sum .lantern/bin/lantern 
e9e3e24fab0824731cf98f5239b5c3af54fbbdf6  .lantern/bin/lantern
```

After killing and restarting the process, Lantern started again with the new version.

Please see my [test .deb package](https://github.com/getlantern/autoupdate/releases/download/2.0.0-beta2/lantern_2.0.0-beta2_amd64.deb) to run the same test.

xiam:
Windows 7: ✓(thanks @myleshorton)

VirtualBox seems to perform better on Linux than on OSX, maybe because Linux leaves a greater amount of RAM available for working.

Well, after a while, the original signature changes to `faf509bfe1da4be94043e99eec378706`, which is expected:

screenshot from 2015-03-29 06 18 26

After a manual restart Lantern still proxies sites without any firewall alert.

screenshot from 2015-03-29 06 13 10

myleshorton:
Awesome @xiam!! So sounds like it's working on all OSes right? Any outstanding issues at this point in your testing?

xiam:
@myleshorton Still can't confirm if working in all OSes, I'm downloading Windows 8, 8.1 and 10 (Yay!) just to be absolutely sure :-).

xiam:
Windows 8: ✓

screenshot from 2015-03-30 10 30 47

xiam:
Windows 8.1: ✓

screenshot from 2015-03-30 10 58 33

xiam:
Windows 10: ✓

screenshot from 2015-03-30 11 25 08

@myleshorton Looks ready! Do you have any other concerns? You mentioned something about the Windows firewall. 

Can you help me reviewing this PR? Thanks.

myleshorton:
Awesome @fffw! I might have to assign this to someone else, unfortunately, as I'll be pretty tied up this week -- will give you details on our call.

atavism:
@myleshorton Happy to look over it after I fix this analytics merge.

myleshorton:
That would be awesome @atavism! Assigning over to you. Thanks.

src/github.com/getlantern/flashlight/autoupdate.go
```diff
@@ -0,0 +1,6 @@
+package main
+
+var (
+	packageVersion   = "2.0.0-beta2"
```
atavism: After an update succeeds, does this need updated?

src/github.com/getlantern/autoupdate-server/server/asset_test.go
```diff
@@ -0,0 +1,16 @@
...
+
+const (
+	testAssetURL = `https://github.com/getlantern/autoupdate-server/releases/download/0.4.0/autoupdate-binary-darwin-amd64.v4`
+)
+
+func TestDownloadAsset(t *testing.T) {
+	if _, err := downloadAsset(testAssetURL); err != nil {
+		t.Fatal(fmt.Errorf("Failed to download asset: %q", err))
+	}
+}
```
atavism: Might want to include the autoupdate packages in testpackages.txt so that we actually run these tests.

src/github.com/getlantern/autoupdate-server/server/asset.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+// downloadAsset grabs the contents of the body of the given URL and stores
+// then into $ASSETS_DIRECTORY/$BASENAME.SHA256_SUM($URL)
+func downloadAsset(uri string) (localfile string, err error) {
+	basename := path.Base(uri)
+
+	if len(basename) > 60 {
+		basename = basename[:60]
+	}
```
atavism: Where does the max length of 60 for the path base come from?

src/github.com/getlantern/autoupdate-server/server/asset_test.go
myleshorton: It might make more sense to make testpackages.txt a blacklist instead of a whitelist -- i.e. don't test these packages that aren't our code. Either that or just test everything in org/getlantern?

atavism: This is our code though, isn't it? I was just including a general comment that we aren't including autoupdate in testpackages.txt

installer-resources/linux/lantern.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
```
xiam: We use this file to bootstrap the Lantern binary in a way that allows the binary to update itself.

src/github.com/getlantern/autoupdate-server/server/asset_test.go
myleshorton: Right. I'm suggesting as a general approach it might make more sense to always test everything in org/getlantern or to use a blacklist. That way we don't have to remember to add something to testpackages.txt every time we add a new package.

atavism: Oh, that makes sense! Maybe we could make that change when we switch our build scripts to use a single Makefile. I think just testing everything under getlantern would work well.

myleshorton: Yeah that makes sense. Then testpackages.txt could just list anything outside of getlantern we want to test, although there probably isn't anything.

xiam: I agree with running all tests that belong to Lantern except a few that might be blacklisted. I'll keep that in mind for the changes I'm proposing to our build scripts.

src/github.com/getlantern/flashlight/autoupdate.go
xiam: Yes. Maybe we should not rely on doing this everytime, this kind of thing could be more easily done by one of our build scripts, in fact I'll add this to my tasks on the build scripts front.

src/github.com/getlantern/autoupdate-server/server/asset.go
xiam: That line looks suspicious indeed! I added a comment to clarify what's the intention of having it.

atavism:
This is working fine for me on OSX and Windows 8.1. I verified the SHA1 checksum changed after I launched Lantern so the update applied successfully. Also tracked progress of the update in the log file:
```
Mar 31 22:46:31.791 - flashlight.autoupdate: Software version: 2.0.0-beta2
Mar 31 22:46:31.791 - autoupdate: Defaulted CheckInterval to 4h0m0s
Mar 31 22:46:32.629 - autoupdate: Attempting to update to 2.0.0-beta3.
...
Mar 31 22:46:41.816 - autoupdate: Patching succeeded!
Mar 31 22:46:41.819 - flashlight.autoupdate: Got update.
```

xiam:
> This is working fine for me on OSX and Windows 8.1. I verified the SHA1 checksum changed after I launched Lantern so the update applied successfully.

@atavism Great! Do you have any other concern you'd like to clarify before merging?

Thanks

atavism:
@xiam Nope, it looks great to me. I still want to try this on Linux myself, but I think we've conducted enough tests already it's fine to merge now!

xiam:
Thank you @atavism!

myleshorton:
Super exciting to have this in guys! Great work @xiam. We'll get this out in a beta 5 shortly. Beta 4 is currently in the QA cycle but should go live fairly soon.

PR #57, opened 2015-02-18
Switched systray to use 2goarray and tarfs to use similar encoding style as 2goarray

oxtoacart:


oxtoacart:
@fffw FYI, I liked 2goarray so much that I retrofitted it.

oxtoacart:
@fffw I ended up reverting this because it slowed down the build a lot (orders of magnitude).  For whatever reason, encoding the bytes as a string allows Go to build much faster than when using a byte array literal.

PR #59, opened 2015-02-18
One way to get public ip

oxtoacart:


src/github.com/getlantern/flashlight/client/balancer.go
```diff
@@ -14,9 +15,18 @@ func (client *Client) initBalancer(cfg *ClientConfig) *balancer.Balancer {
...
+		fd, dialer := s.dialer(cfg.MasqueradeSets)
 		dialers = append(dialers, dialer)
+		if dialer.QOS > highestQOS {
+			highestQOSFrontedDialer = fd
+		}
+	}
+
+	if highestQOSFrontedDialer != nil {
+		go determinePublicIp(highestQOSFrontedDialer)
 	}
```
myleshorton, line 34: Maybe log an error is we don't get a high QoS fronted dialer for any reason?

oxtoacart, line 34: Done.  I made it debug since it doesn't actually prevent the functioning of the system, it just keeps us from getting a public ip.

myleshorton, line 34: Cool thanks. That's certainly true for this iteration, but if/when we add
back in existing functionality that public IP becomes crucial for
determining if you're in a censored region.

On Wed, Feb 18, 2015 at 12:07 PM, oxtoacart notifications@github.com
wrote:

> In src/github.com/getlantern/flashlight/client/balancer.go
> https://github.com/getlantern/flashlight-build/pull/59#discussion_r24935700
> :
> 
> > ```
> >     dialers = append(dialers, dialer)
> > ```
> > -       if dialer.QOS > highestQOS {
> > -           highestQOSFrontedDialer = fd
> > -       }
> > -   }
> >   +
> > -   if highestQOSFrontedDialer != nil {
> > -       go determinePublicIp(highestQOSFrontedDialer)
> >   }
> 
> Done. I made it debug since it doesn't actually prevent the functioning of
> the system, it just keeps us from getting a public ip.
> 
> —
> https://github.com/getlantern/flashlight-build/pull/59/files#r24935700.

## 

## 

President
https://www.getlantern.org

oxtoacart, line 34: Yeah, let's cross that bridge when we come to it.  It's still not 100% clear whether we'll even have the same program for give and get mode.

@xiam This reminds me, we need more than just the public IP.  We need the geolocation information (lat/lon) to place the user's dot on the map.  That information is available in the body of the response.  You can use [this](https://github.com/getlantern/flashlight-build/blob/valencia/src/github.com/getlantern/flashlight/statserver/peer.go#L40) data structure to capture the JSON from the body.  You'll probably want to create a new geo package inside flashlight or something like that and then share it between client and statreporter.

xiam, line 34: @oxtoacart Sounds good, I too think geolocation information could get its own package.

oxtoacart:
Was just a demo, closing.

PR #60, opened 2015-02-19
Enabling CGO on cross-compiles

oxtoacart:
I'm enabling CGO so that we have access to the user HomeDir when cross compiling, as well as getting better DNS performance (theoretically).


oxtoacart:
I just went ahead and merged this, hopefully that's not too controversial :)

PR #62, opened 2015-02-19
websocket branch code review tweaks

oxtoacart:
@atavism I'm providing feedback and tweaks for the websocket branch code review here.  It's not ready yet, but I'll add code and comments as I proceed with my review and let you know when it's ready.


src/github.com/getlantern/flashlight/client/config.go
```diff
@@ -18,9 +18,8 @@ type ClientConfig struct {
 	DumpHeaders    bool // whether or not to dump headers of requests and responses
 	FrontedServers []*FrontedServerInfo
 	ChainedServers map[string]*ChainedServerInfo
+	ProxiedSites   *proxiedsites.Config // List of proxied site domains that get routed through Lantern rather than accessed directly
```
oxtoacart: I moved this so that ProxiedSites comes before MasqueradeSets in the yaml.  It's more interesting and more likely to be edited by the user, so this is a nice convenience.

src/github.com/getlantern/flashlight/config/config.go
```diff
@@ -228,6 +228,11 @@ func (cfg *Config) applyClientDefaults() {
 		}
 	}
 
+	if cfg.Client.ProxiedSites.Cloud == nil || len(cfg.Client.ProxiedSites.Cloud) == 0 {
+		log.Debugf("Loading default cloud proxiedsites")
```
oxtoacart: I'm now defaulting the list of cloud proxiedsites based on a list kept in the binary itself.

oxtoacart: This is just to ensure the best possible out-of-the-box experience.

src/github.com/getlantern/flashlight/genconfig/genconfig.go
```diff
@@ -90,6 +88,11 @@ func main() {
 	if err != nil {
 		log.Fatalf("Unable to format masquerades.go: %s", err)
 	}
+	generateTemplate(model, proxiedSitesTmpl, "../config/proxiedsites.go")
```
oxtoacart, line 91: Generates proxiedsites.go with default list.

src/github.com/getlantern/proxiedsites/proxiedsites.go
```diff
@@ -33,12 +33,12 @@ var (
...
 type Config struct {
-	// Global list of white-listed domains
-	Cloud []string `json:"-"`
-
 	// User customizations
 	Additions []string `json:"Additions, omitempty"`
 	Deletions []string `json:"Deletions, omitempty"`
+
+	// Global list of white-listed domains
+	Cloud []string `json:"-"`
```
oxtoacart: Users are more likely to want to change the Additions and Deletions in the yaml, so I moved Cloud below them to have it appear further down in the YAML.

atavism:
@oxtoacart I know you're in the middle of reviewing changes, but I found a couple things to tweak in the proxied sites implementation (two new error messages + some additional tests I already wrote). I just committed it to this branch.

oxtoacart:
@xiam This still needs to be reviewed, but I think the API won't change very much from this point, so if you want to use this for your geolocation stuff, it's ready.

@atavism This is ready for review.  I refactored a bunch of stuff to achieve a cleaner separation between proxiedsites and flashlight, and within flashlight to achieve a cleaner separation between the general UI code and the proxiedsites-specific functionality.  I also built the default proxied sites list into the binary.

In testing, I see two problems that I'm pretty sure I didn't introduce:
1. When running with the default proxied sites list of 7000 or so sites, making any changes in the textarea is dog slow.  That's because the validation is checking every line on every keystroke.  This is unusable as it stands.
2. When testing with two windows, updates from one window will show up in the other window a few times, but then they stop applying.  I tried fixing this using $apply, which I think helped a little, but it's still inconsistent.

src/github.com/getlantern/flashlight/client/client.go
```diff
@@ -78,4 +78,7 @@ func (client *Client) Configure(cfg *ClientConfig) {
 	client.MinQOS = cfg.MinQOS
 	bal := client.initBalancer(cfg)
 	client.initReverseProxy(bal, cfg.DumpHeaders)
+	client.priorCfg = cfg
+	client.priorTrustedCAs = &x509.CertPool{}
+	*client.priorTrustedCAs = *globals.TrustedCAs
```
oxtoacart, line 83: @atavism This is something that was broken from before your changes.  I noticed that we never stored the old client configuration, so everytime any part of the system configuration changed, the client would reconfigure itself.  This fixes that.

genassets.bash
```diff
@@ -1,3 +1,3 @@
 #!/bin/bash
 
-tarfs -pkg http src/github.com/getlantern/ui/app > src/github.com/getlantern/flashlight/http/resources.go
+tarfs -pkg ui src/github.com/getlantern/ui/app > src/github.com/getlantern/flashlight/ui/resources.go
```
oxtoacart: I renamed http to ui.

src/github.com/getlantern/flashlight/client/config.go
```diff
@@ -19,8 +18,6 @@ type ClientConfig struct {
 	FrontedServers []*FrontedServerInfo
 	ChainedServers map[string]*ChainedServerInfo
 	MasqueradeSets map[string][]*fronted.Masquerade
-
-	ProxiedSites *proxiedsites.Config // List of proxied site domains that get routed through Lantern rather than accessed directly
```
oxtoacart, line 23: I moved proxiedsites.Config from client into the base config.  This is to keep the client from reconfiguring itself needlessly, since it actually doesn't care about the proxied sites.

src/github.com/getlantern/flashlight/config/config.go
```diff
@@ -205,6 +210,22 @@ func (cfg *Config) ApplyDefaults() {
...
+				Additions: []string{},
+				Deletions: []string{},
+			},
+			Cloud: []string{},
+		}
+	}
+
+	if cfg.ProxiedSites.Cloud == nil || len(cfg.ProxiedSites.Cloud) == 0 {
+		log.Debugf("Loading default cloud proxiedsites")
+		cfg.ProxiedSites.Cloud = defaultProxiedSites
```
oxtoacart, line 224: Here is where we default the proxiedsites list to the default list (generated by genconfig).

src/github.com/getlantern/flashlight/proxiedsites/proxiedsites.go
```diff
@@ -0,0 +1,70 @@
...
+			log.Errorf("Unable to publish delta to UI: %v", err)
+		}
+		uichannel.Out <- b
+	}
+	startMutex.Unlock()
+}
+
+func start() {
+	// Register the PAC handler
+	url := ui.Handle("/proxy_on.pac", http.HandlerFunc(proxiedsites.ServePAC))
```
oxtoacart, line 41: I didn't see any reason to store the pac file on disk, and it added a fair amount of complexity, so now the proxiedsites package just serves up the PAC directly.

src/github.com/getlantern/flashlight/proxiedsites/proxiedsites.go
```diff
@@ -0,0 +1,70 @@
...
+	log.Debugf("Serving PAC file at %v", url)
+
+	// Establish a channel to the UI for sending and receiving updates
+	uichannel = ui.NewChannel("/data", func(write func([]byte) error) error {
```
oxtoacart, line 45: @atavism @xiam UIChannel is an abstraction that I added to represent a channel of communication to the UI.  All browser windows are represented by this channel, so that when any of them write, we get the message, and if we write to the channel, all of the browsers get the message.  I use this abstraction here for the proxiedsites functionality, and @xiam can use it elsewhere for the geoip functionality (just use a different path than `/data`).

src/github.com/getlantern/proxiedsites/proxiedsites.go
```diff
@@ -1,269 +1,217 @@
-// package proxiedsites is a module used to manage the list of sites
-// being proxied by Lantern
-// when the list is modified using the Lantern UI, it propagates
-// to the default YAML and PAC file configurations
+// proxiedsites manages a list of proxied sites, including a default set of
+// sites (cloud) and user-applied customizations to that list. It provides an
+// implementation of the http.Handler interface that serves up a PAC file based
+// on the currently active proxied sites (cloud + additions - deletions).
 package proxiedsites
```
oxtoacart, line 5: proxiedsites was simplified considerably.
1. It's all just package level functions now (no ProxiedSites data structure)
2. There's no pac file, it's just bytes in memory. ProxiedSites provides a function that can be used as an HTTP handler that serves up the pac file from memory.
3. Additions and Deletions were factored into a Delta data structure since these are sometimes sent independently of the Global list.
4. It follows the usual Configure() paradigm.  Also, by having Configure() return a Delta representing the changes, we no longer need any channels in this package.

src/github.com/getlantern/ui/app/js/app.js
```diff
@@ -73,25 +73,35 @@ var app = angular.module('app', [
...
-            var entries = $rootScope.entries;
-            entries.push(msg.Additions);
-
-            $rootScope.entries = entries;
-            $rootScope.originalList = entries;
+            var entries = $rootScope.entries.slice(0);
+            if (msg.Additions) {
+              entries = _.union(entries, msg.Additions);
+            }
+            if (msg.Deletions) {
```
oxtoacart, line 92: The original code was not applying deletions.  Now it does.

src/github.com/getlantern/ui/app/js/app.js
```diff
@@ -73,25 +73,35 @@ var app = angular.module('app', [
...
+              entries = _.difference(entries, msg.Deletions)
+            }
+            entries = _.compact(entries);
```
oxtoacart, line 95: This removes blank lines.

src/github.com/getlantern/yamlconf/yamlconf.go
```diff
@@ -211,7 +211,7 @@ func (m *Manager) processUpdates() {
 		select {
 		case delta := <-m.deltasCh:
 			log.Trace("Apply delta")
-			updated, err := m.copy(m.cfg)
+			updated, err := m.copy(m.getCfg())
```
oxtoacart, line 214: This was a little fix for a potential concurrency issue on accessing m.cfg.

src/github.com/gorilla/context/.travis.yml
```diff
@@ -0,0 +1,9 @@
+language: go
```
oxtoacart, line 1: I vendored gorilla.

src/github.com/robertkrimen/otto/.gitignore
```diff
@@ -1,5 +0,0 @@
-/.test
```
oxtoacart, line 1: We don't need otto anymore since we're not parsing a PAC file.

xiam:
Thank you @oxtoacart, I'm working now on the websocket-ox-tweaks branch.

src/github.com/getlantern/flashlight/proxiedsites/proxiedsites.go
xiam, line 45: Got it. I'll need to send a message to `Out`, right?

oxtoacart, line 45: Yes, like [here](https://github.com/getlantern/flashlight-build/pull/62/files#diff-c117cc8455c34de273b2b92fc7c9d133R33).

src/github.com/getlantern/flashlight/ui/ws.go
```diff
@@ -0,0 +1,153 @@
...
+		log.Tracef("About to read from connection to %v", c.URL)
+		conn.read()
+	})
+
+	return c
+}
+
+func newUIChannel(url string) *UIChannel {
+	in := make(chan []byte, 100)
+	out := make(chan []byte)
```
xiam, line 92: I noticed that the out channel is not buffered (`newUIChannel()`), so a goroutine that writes to it would block until the message is received on the client by WriteMessage. Since we're already thinking on a situation when two or more browsers are opened it would probably be better to have a small buffer here.

oxtoacart, line 92: Maybe.  Keep in mind that the network layer is buffering stuff too, so this isn't actually blocking until the message has been received by the client, it's just blocking until the data has been written to the network buffers.

xiam:
It looks like we're assuming we'll send only one kind of message through the proxy (data about proxied sites): https://github.com/getlantern/flashlight-build/blob/websocket-ox-tweaks/src/github.com/getlantern/ui/app/js/app.js#L80

I'm I understanding that right?

So right now I can't use the same onMessage to receive a message with the user's geolocation data.

To fix that, I'm thinking on creating a simple envelope like:

``` json
{
 "MessageType": "ProxiedSites",
 "Data": { /* original data */ },
}
```

I would use another MessageType for the user's geolocation information and handle it in the same onMessage.

And in order to do that, I just have to add an `Envelope` struct to the `ws.go` file and update a few lines in `proxiedsites.go` and in `app.js`.

Do you have another suggestion?

atavism:
> And in order to do that, I just have to add an Envelope struct to the ws.go file and update a few lines in proxiedsites.go and in app.js.

Data was intended to support multiple message types anyway so that sounds like a good change to make.

oxtoacart:
@xiam Actually the whole reason that I introduced the UIChannel abstraction is so that you can set up a separate channel for things other than proxiedsites.  My intent was not for you to use the same UIChannel as proxiedsites.

atavism:
@oxtoacart Who did you want to review these changes?

xiam:
@oxtoacart What if I leave /data as a reserved websocket for proxiedsites and create a new one to handle miscellaneous messages? The user's geolocation information could be the first of those messages and we'd leave the implementation ready for any future usage, such as reporting errors, etc.

In the other hand, implementing everything in `/data` would have the additional benefit of having only one websocket connection opened (just one error management routine and one point of failure less to worry about).

atavism:
@oxtoacart Are changes to the UI code no longer automatically detected? It looks like I have to restart flashlight each time now.

oxtoacart:
> Who did you want to review these changes?

@atavism You.  Just assigned it to you.

oxtoacart:
> Are changes to the UI code no longer automatically picked up? It looks like I have to restart flashlight each time now.

How are you running flashlight?

src/github.com/getlantern/flashlight/ui/ui.go
```diff
@@ -0,0 +1,72 @@
...
+	"path"
+	"path/filepath"
+
+	"github.com/getlantern/golog"
+	"github.com/getlantern/tarfs"
+	"github.com/skratchdot/open-golang/open"
+)
+
+const (
+	localResourcesPath = "../ui/app"
```
oxtoacart: @atavism UI resources are specified using a relative path that assumes that you're running flashlight from within the flashlight folder during development.

atavism: I've been running flashlight from the flashlight-build directory. So this is no longer looking for a ui directory in `src/github.com/getlantern`? Running flashlight from `src/github.com/getlantern/flashlight` I'm seeing 

> go run flashlight.go -role client -addr localhost:10080 -uiaddr localhost:16785 -cloudconfig http://s3.amazonaws.com/lantern_config/cloud.valencia.yaml.gz
> ./flashlight.go:228: undefined: Asset

oxtoacart:
@xiam 

> What if I leave /data as a reserved websocket for proxiedsites and create a new one to handle miscellaneous messages? The user's geolocation information could be the first of those messages and we'd leave the implementation ready for any future usage, such as reporting errors, etc.

Maybe.  It'll be clearer once we have requirements for other data.

> In the other hand, implementing everything in /data would have the additional benefit of having only one websocket connection opened (just one error management routine and one point of failure less to worry about).

These are websockets to a local server, so I wouldn't worry about failures.  In terms of error handling, the only handling that we do is reconnecting.  That can be gotten for free by using a library like this one - https://github.com/joewalnes/reconnecting-websocket

Also, if we ever have application-specific error handling, it will probably vary by channel.  So for example, if we can't get your authentication information, that might cause us to actually show an error, whereas if we can't get the latest stats for peers we may just ignore that.

In general, I like the pattern of dedicated channels because it avoids having giant `switch (messageType)` statements everywhere.  It's possible to build a nice API like that and multiplex messages over a single websocket, but I don't know that there's much benefit to that.  And the benefit to giving everything its own websocket is that a slow reader in one subsystem doesn't hold up processing in another.

Just my 2 cents.

src/github.com/getlantern/proxiedsites/proxiedsites.go
atavism, line 5: Where is the flashlight.yaml file now? Is that no longer being generated? I use to be able to update the list of proxied sites directly from there. How do I make on-the-fly changes to the list now without going through the UI?

atavism:
> In general, I like the pattern of dedicated channels because it avoids having giant switch (messageType) statements everywhere. It's possible to build a nice API like that and multiplex messages over a single websocket, but I don't know that there's much benefit to that. And the benefit to giving everything its own websocket is that a slow reader in one subsystem doesn't hold up processing in another.

We didn't really get to discuss the merits of both approaches at all. My original implementation used a single websocket for the reasons already outlined. Nothing about the previous API necessitated using a giant switch statement: we could have configured the mux router to use multiple HTTP handlers with a single websocket. The issue of a slow reader could have been handled across the board by adding a read timeout to the processing time of every request. There was some error handling already that's since been removed.

src/github.com/getlantern/flashlight/ui/ui.go
myleshorton: @atavism for whatever reason I see the same thing when running with `go run` but not with the precompiled ./flashlight. So this works for me:

```
./flashlight -role client -addr localhost:10080 -uiaddr localhost:16785 -cloudconfig http://s3.amazonaws.com/lantern_config/cloud.valencia.yaml.gz
```

oxtoacart:
@atavism 

> I've been running flashlight from the flashlight-build directory. So this is no longer looking for a ui directory in `src/github.com/getlantern`?

Yeah, sorry, I changed that. I've updated it now to make the path relative to the GOPATH so it should work no matter where you run flashlight.

> Running flashlight from `src/github.com/getlantern/flashlight` I'm seeing 
> 
> > go run flashlight.go -role client -addr localhost:10080 -uiaddr localhost:16785 -cloudconfig http://s3.amazonaws.com/lantern_config/cloud.valencia.yaml.gz
> > ./flashlight.go:228: undefined: Asset

Very strange.  Do you have the file icons.go in the same folder as flashlight.go?

src/github.com/getlantern/proxiedsites/proxiedsites.go
oxtoacart, line 5: It's in ~/Library/Application Support/Lantern/lantern.yaml

atavism:
> Very strange. Do you have the file icons.go in the same folder as flashlight.go?

Yes.

oxtoacart:
> We didn't really get to discuss the merits of both approaches at all.

We're discussing it now.

> My original implementation used a single websocket for the reasons already outlined. Nothing about the previous API necessitated using a giant switch statement: we could have configured the mux router to use multiple HTTP handlers with a single websocket.

Can you show me how please (code sample, reference link)?  If we can make this work, we should in theory be able to keep the UIChannel API on the server and just use a single websocket.

> The issue of a slow reader could have been handled across the board by adding a read timeout to the processing time of every request.

Not really.  What I'm talking about is that if we have multiple consumers of messages from the websocket, the read loop will be dispatching to those consumers.  If one of those consumers is slow in doing its work, it will hold up subsequent reads.  We could introduce additional channels for each consumer to mitigate this, but that adds complexity.

> There was some error handling already that's since been removed.

Sorry about that.  Can you point me to the original code please?

oxtoacart:
@atavism 

> > Very strange. Do you have the file icons.go in the same folder as flashlight.go?
> 
> Yes.

Well, that's where the Asset function is defined.  Can you please check if your icons.go has the Asset function?  You can also try doing a full rebuild with `go install -a` to see if that clears things up.

BTW, can you please copy and paste the value of your `$GOPATH`?

src/github.com/getlantern/flashlight/ui/ui.go
oxtoacart: @myleshorton Just changed this to make it relative to the GOPATH so it should work no matter where you run flashlight.  Can you please try again?

atavism:
```
➜  flashlight git:(websocket-ox-tweaks) ✗ go run flashlight.go -role client -addr localhost:10080 -uiaddr localhost:16785 -cloudconfig http://s3.amazonaws.com/lantern_config/cloud.valencia.yaml.gz
# command-line-arguments
./flashlight.go:228: undefined: Asset
➜  flashlight git:(websocket-ox-tweaks) ✗ echo $GOPATH
/Users/todd/Dev/go/src/github.com/getlantern/flashlight-build
```

In icons.go:

```
   // Asset loads and returns the asset for the given name.
   // It returns an error if the asset could not be found or
   // could not be loaded.
   func Asset(name string) ([]byte, error) {
       cannonicalName := strings.Replace(name, "\\", "/", -1)
       if f, ok := _bindata[cannonicalName]; ok {
           return f()
       }
       return nil, fmt.Errorf("Asset %s not found", name)
   }
```

oxtoacart:
`go run flashlight.go`

That ignores all the other go files.  Try this:

`go install && flashlight`

src/github.com/getlantern/flashlight/ui/ui.go
myleshorton: I think this was intended for @atavism

On Fri, Feb 20, 2015 at 10:22 AM, oxtoacart notifications@github.com
wrote:

> In src/github.com/getlantern/flashlight/ui/ui.go
> https://github.com/getlantern/flashlight-build/pull/62#discussion_r25090293
> :
> 
> > +package ui
> > +
> > +import (
> > -   "fmt"
> > -   "net"
> > -   "net/http"
> > -   "path"
> > -   "path/filepath"
> >   +
> > -   "github.com/getlantern/golog"
> > -   "github.com/getlantern/tarfs"
> > -   "github.com/skratchdot/open-golang/open"
> >   +)
> >   +
> >   +const (
> > -   localResourcesPath = "../ui/app"
> 
> @myleshorton https://github.com/myleshorton Just changed this to make
> it relative to the GOPATH so it should work no matter where you run
> flashlight. Can you please try again?
> 
> —
> https://github.com/getlantern/flashlight-build/pull/62/files#r25090293.

## 

## 

President
https://www.getlantern.org

myleshorton:
> go run flashlight.go
> 
> That ignores all the other go files. Try this:
> 
> go install && flashlight

Oh right -- that's what I was saying worked for me (well, go build && ./flashlight), but I didn't fully get why. I thought go run xxx.go would pull in dependencies and build them, but I guess not. This is clear with:

```
$ flashlight (websocket-ox-tweaks)]$ go build flashlight.go
# command-line-arguments
./flashlight.go:228: undefined: Asset
```

src/github.com/getlantern/proxiedsites/proxiedsites.go
oxtoacart, line 5: BTW, I try to be good about logging stuff like this, so if you're ever looking to see where stuff is coming from, the log output is useful:

```
flashlight: Placing logs in /Users/ox.to.a.cart/Library/Logs/Lantern
Feb 20 17:56:27.919 - flashlight: ---- flashlight version development (now) ----
Feb 20 17:56:27.919 - flashlight.config: Placing configuration in /Users/ox.to.a.cart/Library/Application Support/Lantern
```

src/github.com/getlantern/flashlight/ui/ui.go
```diff
@@ -0,0 +1,72 @@
...
+	"github.com/getlantern/tarfs"
+	"github.com/skratchdot/open-golang/open"
+)
+
+const ()
+
+var (
+	log = golog.LoggerFor("ui")
+
+	localResourcesPath = filepath.Join(os.Getenv("GOPATH"), "src/github.com/getlantern/ui/app")
```
atavism: The code I had before actually looked for the ui directory relative to the source file of the function being invoked which conveniently handled most use cases. It used [runtme.Caller](http://golang.org/pkg/runtime/#Caller)

src/github.com/getlantern/flashlight/ui/ws.go
```diff
@@ -0,0 +1,153 @@
...
+	go c.write()
+	return c
+}
+
+func (c *UIChannel) write() {
+	defer func() {
+		log.Tracef("Closing all websockets to %v", c.URL)
+		c.m.Lock()
+		for _, conn := range c.conns {
+			conn.ws.Close()
```
atavism, line 113: What about deleting the connection ID from the map?

atavism, line 113: Oh, this closes every connection. Nevermind.

src/github.com/getlantern/proxiedsites/proxiedsites.go
```diff
@@ -1,269 +1,218 @@
...
+func toStrings(s set.Interface) []string {
+	sl := s.List()
+	l := make([]string, len(sl))
+	for i, s := range sl {
+		l[i] = s.(string)
 	}
-	go ps.updatePacFile()
-	return ps
+	sort.Strings(l)
+	return l
```
atavism, line 99: Why use this custom implementation over the existing set.StringSlice? Because it sorts at the end? It would help to at least mention that in the comment.

src/github.com/getlantern/proxiedsites/proxiedsites.go
```diff
@@ -1,269 +1,218 @@
...
 
-func SetPacFile(pacFile string) {
-	PacFilePath = pacFile
+// ServePAC serves up the PAC file and can be used as an http.HandlerFunc
+func ServePAC(resp http.ResponseWriter, req *http.Request) {
+	resp.Header().Set("Content-Type", "application/x-ns-proxy-autoconfig")
+	resp.WriteHeader(http.StatusOK)
+	cfgMutex.RLock()
+	resp.Write([]byte(pacFile))
+	cfgMutex.RUnlock()
```
atavism, line 175: Nice. Definitely makes sense to store it in-memory and skip saving a file entirely since we already write the latest changes to the YAML configuration.

src/github.com/robertkrimen/otto/.gitignore
atavism, line 1: Right. I was just using it previously to test the validity of the updated PAC file.

src/github.com/getlantern/flashlight/ui/ui.go
oxtoacart: Ahh, cool.  Sorry about that - restored now.  I think I missed it because I ditched the branch that bypassed tarfs.

src/github.com/getlantern/flashlight/ui/ws.go
oxtoacart, line 113: You're right, I should delete them.  Fixed.  Thanks!

src/github.com/getlantern/proxiedsites/proxiedsites.go
oxtoacart, line 99: Oh hey cool, that saves some code.  Fixed!

src/github.com/getlantern/proxiedsites/proxiedsites.go
oxtoacart, line 175: Thanks.

src/github.com/robertkrimen/otto/.gitignore
oxtoacart, line 1: Ah, my bad.  I saw it in proxiedsites.go rather than proxiedsites_test.go and assumed it was part of the production code.

oxtoacart, line 1: For future reference, it's nice to put testing code in the _test.go files, which keeps it from getting compiled into the main app.

src/github.com/getlantern/flashlight/ui/ws.go
```diff
@@ -0,0 +1,153 @@
...
+				return
+			}
+		}
+		c.nextId += 1
+		conn := &wsconn{
+			id: c.nextId,
+			c:  c,
+			ws: ws,
+		}
+		c.conns[conn.id] = conn
```
atavism, line 81: What purpose is the connection ID actually serving here? You just ignore the ID everywhere you iterate through the connections anyway.

src/github.com/getlantern/flashlight/ui/ws.go
```diff
@@ -0,0 +1,153 @@
...
+	c  *UIChannel
+	ws *websocket.Conn
+}
+
+func (c *wsconn) read() {
+	for {
+		_, b, err := c.ws.ReadMessage()
+		if err != nil {
+			if err != io.EOF {
+				log.Debugf("Error reading from UI: %v", err)
```
atavism: Not really a Debug message.

src/github.com/getlantern/flashlight/ui/ws.go
oxtoacart, line 81: I use it to delete connections from the map of connections upon closing them.

src/github.com/getlantern/flashlight/ui/ws.go
oxtoacart: Fixed.

src/github.com/getlantern/flashlight/ui/ui.go
```diff
@@ -0,0 +1,72 @@
...
+		Handler: r,
+	}
+	go func() {
+		err := server.Serve(l)
+		if err != nil {
+			log.Errorf("Error serving: %v", err)
+		}
+	}()
+	uiaddr = fmt.Sprintf("http://%v", l.Addr().String())
+	log.Debugf("UI available at %v", uiaddr)
```
atavism, line 77: Do we want to continue reporting the UI is available if Serve() fails? I'd just move that debugging info after the nil error check (and return) inside the goroutine.

oxtoacart, line 77: I can't do that because Serve blocks until the server is done serving.

oxtoacart:
@atavism Thanks for reviewing!  I hope that the modified structure makes sense and that I wasn't too gratuitous in my tweaks.

PR #63, opened 2015-02-20
always proxy fetching of config updates when running as client

aranhoide:
I've tested this change in devel by running both as client and server and checking that the following gets printed only in the first case.

`balancer: Dialing tcp://s3.amazonaws.com:443 with fronted proxy at fallbacks.getiantem.org:443 using masquerade set cloudflare`

I also run with TRACE=true and grepped for doFetchCloudConfig.

I couldn't build the valencia branch in first place, apparently for reasons that have nothing to do with this change:

```
# pkg-config --cflags gtk+-3.0 appindicator3-0.1
Package gtk+-3.0 was not found in the pkg-config search path.
Perhaps you should add the directory containing `gtk+-3.0.pc'
to the PKG_CONFIG_PATH environment variable
No package 'gtk+-3.0' found
Package appindicator3-0.1 was not found in the pkg-config search path.
Perhaps you should add the directory containing `appindicator3-0.1.pc'
to the PKG_CONFIG_PATH environment variable
No package 'appindicator3-0.1' found
pkg-config: exit status 1
```

### DIFF
--- a/src/github.com/getlantern/flashlight/config/config.go
+++ b/src/github.com/getlantern/flashlight/config/config.go
@@ -273,21 +273,28 @@ func (cfg Config) cloudPollSleepTime() time.Duration {
 	return time.Duration((CloudConfigPollInterval.Nanoseconds() / 2) + rand.Int63n(CloudConfigPollInterval.Nanoseconds()))
 }
 
-func (cfg Config) fetchCloudConfig() ([]byte, error) {
+func (cfg Config) fetchCloudConfig() (bytes []byte, err error) {
 	log.Debugf("Fetching cloud config from: %s", cfg.CloudConfig)
-	// Try it unproxied first
-	bytes, err := cfg.doFetchCloudConfig("")
-	if err != nil && cfg.IsDownstream() {
-		// If that failed, try it proxied
-		bytes, err = cfg.doFetchCloudConfig(cfg.Addr)
+
+	if cfg.IsDownstream() {
+		// Clients must always proxy the request
+		if cfg.Addr == "" {
+			err = fmt.Errorf("No proxyAddr")
+		} else {
+			bytes, err = cfg.doFetchCloudConfig(cfg.Addr)
+		}
+	} else {
+		bytes, err = cfg.doFetchCloudConfig("")
 	}
 	if err != nil {
-		return nil, fmt.Errorf("Unable to read yaml from %s: %s", cfg.CloudConfig, err)
+		bytes = nil
+		err = fmt.Errorf("Unable to read yaml from %s: %s", cfg.CloudConfig, err)
 	}
-	return bytes, err
+	return
 }
 
 func (cfg Config) doFetchCloudConfig(proxyAddr string) ([]byte, error) {
+	log.Tracef("doFetchCloudConfig via '%s'", proxyAddr)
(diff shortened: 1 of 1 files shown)


src/github.com/getlantern/flashlight/config/config.go
```diff
@@ -273,21 +273,28 @@ func (cfg Config) cloudPollSleepTime() time.Duration {
...
 	log.Debugf("Fetching cloud config from: %s", cfg.CloudConfig)
-	// Try it unproxied first
-	bytes, err := cfg.doFetchCloudConfig("")
-	if err != nil && cfg.IsDownstream() {
-		// If that failed, try it proxied
-		bytes, err = cfg.doFetchCloudConfig(cfg.Addr)
+
+	if cfg.IsDownstream() {
+		// Clients must always proxy the request
+		if cfg.Addr == "" {
```
aranhoide, line 281: Just being extra paranoid here, since addr is mandatory I think.  I'll take this check out if you prefer.

xiam, line 281: I think that extra check is actually a good addition, it clearly says "we're not taking any chance of passing doFetchCloudConfig an empty string as argument".

xiam:
Cool @aranhoide, looks fine here. Thanks!

> I couldn't build the valencia branch in first place, apparently for reasons that have nothing to do with this change

That looks like the problem I got the other day with systray, @oxtoacart told me he's already aware of that, nothing to worry about.

@oxtoacart This was assigned to you but since this closes a P1 I took the liberty of reviewing it, I did and I think we can merge it right away.

I'll leave the actual merge to you, though. Just in case you want to check it again.

oxtoacart:
@aranhoide There are now some native build prerequisites for building on Linux.  I've added some information to the README - https://github.com/getlantern/flashlight-build/commit/d2ce30e342c61d5ecc38a9d5ecfcd3761743dded

src/github.com/getlantern/flashlight/config/config.go
oxtoacart, line 281: I think this is a good thing to check too, but it might be better checked here - https://github.com/getlantern/flashlight-build/blob/valencia/src/github.com/getlantern/flashlight/config/config.go#L79

That way, we don't even bother polling, as opposed to continuously polling and failing.

oxtoacart:
LGTM other than the nit on checking for an empty address. I'll leave it up to @aranhoide if you want to change that.

src/github.com/getlantern/flashlight/config/config.go
aranhoide, line 281: IIUC we are not "polling" anyway, in the sense that no connections are attempted.  So the only effect is negligible CPU usage.

If an empty address was expected at all in normal usage, I'd probably make the change you suggest.  Since this is a defensive check, in principle I prefer to keep it close to where the sensitive actions (in this case, sending out requests to S3) are actually performed, so more possible causes for an empty address are caught.  So, if anything, my inclination woud be to move this into doFetchCloudConfig.

But I'm not convinced of that either, and considering that the importance of this check is dubious to begin with, I don't think it's worth holding this PR to pinpoint the perfect location for it.

aranhoide:
As I replied in that thread, I don't think the empty address check merits holding this PR either way, so I'm merging this.

PR #64, opened 2015-02-20
send error logs to loggly

fffw:
I drafted simple remote logging and tested on Mac, you can see them by filter osName as "darwin" at https://lantern.loggly.com. Log didn't get written on Windows, even to local files. I'm checking.

It's modeled after [Lantern java](https://github.com/getlantern/lantern/blob/devel/src/main/java/org/lantern/loggly/LogglyAppender.java#L57) but left some fields empty or removed.

This resolves https://github.com/getlantern/lantern/issues/2232


fffw:
Oh, `io.MultiWriter` will stop writing when any error encountered. On Windows, `Stderr` will can't be written to. Have to user our own version.

oxtoacart:
This is a good start.  I just have a couple of comments.

src/github.com/getlantern/flashlight/logging.go
```diff
@@ -0,0 +1,102 @@
...
+		return fmt.Fprintf(w, "%s - ", time.Now().In(time.UTC).Format(LogTimestampFormat))
+	})
+}
+
+type logglyErrorWriter struct {
+	l *loggly.Client
+}
+
+func (w logglyErrorWriter) Write(b []byte) (int, error) {
+	return writeToLoggly(w.l, "ERROR", string(b))
```
oxtoacart: I assume that we want to treat lines as messages, since that's how we're logging things.  This treats buffers as messages.  There's no guarantee that `b` will contain a whole line, nor is there any guarantee that it will contain only one line.  You'll need to break out lines, similarly to the way `wfilter` used to do.

src/github.com/segmentio/go-loggly
```diff
@@ -0,0 +1 @@
+Subproject commit e78f6971ebca5835614673e2f5f6a47ca5f13501
```
oxtoacart: We don't want submodules.  Did you `go get` these?  If so, please delete and use `gost get` instead.  See the [Gost README](https://github.com/getlantern/gost) for information.

src/github.com/getlantern/flashlight/logging.go
fffw: This again is caused by mixing different layer of abstraction. I think the correct way is to make sure each log entry is written in whole, rather than patch at an inappropriate byte stream layer. 

As I'm planing a more reliable way to write to Loggly, is it ok to leave it as is if we do need it in Valencia release?

src/github.com/getlantern/flashlight/logging.go
```diff
@@ -0,0 +1,104 @@
...
+}
+
+func writeToLoggly(l *loggly.Client, level string, msg string) (int, error) {
+	extra := map[string]string{
+		"logLevel":  level,
+		"osName":    runtime.GOOS,
+		"osArch":    runtime.GOARCH,
+		"osVersion": "",
+		"language":  "",
```
oxtoacart: I've integrated your i18n library, so you should be able to get language.

src/github.com/getlantern/flashlight/logging.go
```diff
@@ -0,0 +1,104 @@
...
+func (w logglyErrorWriter) Write(b []byte) (int, error) {
+	return writeToLoggly(w.l, "ERROR", string(b))
+}
+
+func writeToLoggly(l *loggly.Client, level string, msg string) (int, error) {
+	extra := map[string]string{
+		"logLevel":  level,
+		"osName":    runtime.GOOS,
+		"osArch":    runtime.GOARCH,
+		"osVersion": "",
```
oxtoacart: This thread has some insights into getting the osVersion - https://groups.google.com/forum/#!topic/golang-nuts/NzEPGqlHIYQ

src/github.com/getlantern/flashlight/logging.go
```diff
@@ -0,0 +1,104 @@
...
+		"language":  "",
+		"country":   "",
+		"timeZone":  "",
```
oxtoacart: Time zone should be available from time.Local.

src/github.com/getlantern/flashlight/logging.go
```diff
@@ -0,0 +1,104 @@
...
+		"version":   version,
```
oxtoacart: Might as well add build date too (you could just include that with version like `fmt.Sprintf("%v (%v)", version, buildDate`.

src/github.com/getlantern/flashlight/logging.go
```diff
@@ -0,0 +1,104 @@
...
+	"github.com/dogenzaka/rotator"
+	"github.com/getlantern/appdir"
+	"github.com/getlantern/golog"
+	"github.com/getlantern/wfilter"
+	"github.com/segmentio/go-loggly"
+)
+
+var (
+	LogTimestampFormat = "Jan 02 15:04:05.000"
+	logglyToken        = "469973d5-6eaf-445a-be71-cf27141316a1"
```
oxtoacart: I think we should make this something that's passed in at compile time, much like version and buildDate (see crosscompile.bash).  The reason is that if someone forks flashlight, we don't want them reporting to the same Loggly account that we use.

src/github.com/getlantern/flashlight/logging.go
oxtoacart: BTW, if this is too much work, I suggest that you just open a new ticket for it and don't worry about it now.

src/github.com/getlantern/flashlight/logging.go
```diff
@@ -0,0 +1,104 @@
...
+func (t *nonStopWriter) Write(p []byte) (n int, err error) {
+	for _, w := range t.writers {
+		n, _ = w.Write(p)
+	}
+	return len(p), nil
+}
+
+// NonStopWriter creates a writer that duplicates its writes to all the
+// provided writers, similar to the Unix tee(1) command.
+func NonStopWriter(writers ...io.Writer) io.Writer {
```
oxtoacart: This is really an improved version of io.MultiWriter, right?  Might want to add that to the comment.

src/github.com/getlantern/flashlight/logging.go
fffw: It requires time calculation. I prefer just leave it to be detected at os level.

fffw: Oh, I just noticed that it's a variable! Thanks @oxtoacart

src/github.com/getlantern/flashlight/logging.go
```diff
@@ -0,0 +1,115 @@
...
+	logglyToken        string
+	versionToLoggly    = fmt.Sprintf("%v (%v)", version, buildDate)
+	lang               string
+	country            string
+	tz                 string
+)
+
+func init() {
+	lang, _ = jibber_jabber.DetectLanguage()
+	country, _ = jibber_jabber.DetectTerritory()
```
oxtoacart: I think we should use the geoip information from @xiam instead of this.  DetectTerritory is looking at language settings, whereas what we want is actual physical location.

src/github.com/getlantern/flashlight/logging.go
oxtoacart: I just realized that you will need to buffer this now.  The reason is that writing to Loggly involves the network and will block for potentially long periods of time.  This means that when code calls log.Error(), that call will potentially block for a long time.  That's not acceptable, since that call can appear on performance critical paths.  Introducing a buffer and doing the Loggly reporting on a goroutine would fix the problem.

fffw: Sorry I have no access to PC currently.

go-loggly is buffered and async itself.

src/github.com/getlantern/flashlight/logging.go
fffw: Oh I'll take a look at GeoIP, thanks.

src/github.com/getlantern/flashlight/logging.go
oxtoacart: Awesome, thanks!

src/github.com/getlantern/flashlight/logging.go
oxtoacart: NP.  The geoip stuff hasn't been merged yet, so I'm thinking let's go ahead and merge your stuff and then add the country info once geoip has been merged.  What do you think?

src/github.com/getlantern/flashlight/logging.go
oxtoacart: Thanks for opening the ticket.

oxtoacart:
LGTM

oxtoacart:
@fffw I went ahead and merged and made a note in the ticket that we still need to integrate the geoip stuff.

fffw:
@oxtoacart I completely agree. Thank you.

PR #65, opened 2015-02-20
Integrate i18n into flashlight-build

oxtoacart:


src/github.com/getlantern/i18n/.gitignore
```diff
@@ -0,0 +1,24 @@
+# Compiled Object files, Static and Dynamic libs (Shared Objects)
```
oxtoacart, line 1: The i18n code has already been reviewed, this just pulls it into flashlight-build.

src/github.com/getlantern/jibber_jabber/.travis.yml
```diff
@@ -0,0 +1,11 @@
+language: go
```
oxtoacart, line 1: jibber_jabber is a transitive dependency of i18n, no review needed.

src/github.com/getlantern/tarfs/tarfs.go
```diff
@@ -91,6 +91,26 @@ func New(tarData []byte, local string) (*FileSystem, error) {
 	return fs, nil
 }
 
+// SubDir returns a FileSystem corresponding to the given directory in the
+// original FileSystem.
+func (fs *FileSystem) SubDir(dir string) *FileSystem {
```
oxtoacart, line 96: New tarfs function to slice off a subdirectory as its own FileSystem.

src/github.com/getlantern/flashlight/flashlight.go
```diff
@@ -231,9 +243,14 @@ func configureSystemTray() {
 	}
 	systray.SetIcon(icon)
 	systray.SetTooltip("Lantern")
-	quit := systray.AddMenuItem("Quit", "Quit Lantern")
+	show := systray.AddMenuItem(i18n.T("TRAY_SHOW_LANTERN"), i18n.T("SHOW"))
+	quit := systray.AddMenuItem(i18n.T("TRAY_QUIT"), i18n.T("QUIT"))
 	go func() {
-		<-quit.ClickedCh
-		os.Exit(0)
+		select {
```
myleshorton: This needs to happen inside a loop, otherwise only one command will ever be executed!

myleshorton:
We also used to just use the translation files from lantern-ui, which is handy because then translators just have to worry about that single file when translating the Lantern client. Looks like it should be pretty easy to just set the directory to the same directory the UI is using, however, and to use the same format?

src/github.com/getlantern/flashlight/flashlight.go
oxtoacart: Good catch, fixed!

oxtoacart:
> We also used to just use the translation files from lantern-ui, which is handy because then translators just have to worry about that single file when translating the Lantern client. Looks like it should be pretty easy to just set the directory to the same directory the UI is using, however, and to use the same format?

We still use the translation files from lantern-ui - https://github.com/getlantern/flashlight-build/blob/issue2240-merge/src/github.com/getlantern/flashlight/ui/ui.go#L32

myleshorton:
> We still use the translation files from lantern-ui - https://github.com/getlantern/flashlight-build/blob/issue2240-merge/src/github.com/getlantern/flashlight/ui/ui.go#L32

Awesome!

src/github.com/getlantern/flashlight/flashlight.go
myleshorton: I don't see that fix. Maybe on another branch?

myleshorton:
Also, the windows version of this is a little wonky, as it flashes the command prompt window each time you show Lantern. I can't remember offhand how to avoid that but will look. I think we're using some version of `start` from the command prompt to trigger that right? There's some tweak to be made there.

src/github.com/getlantern/flashlight/flashlight.go
oxtoacart: Oops, too many branches!  Fixed in here now.

oxtoacart:
> Also, the windows version of this is a little wonky, as it flashes the command prompt window each time you show Lantern. I can't remember offhand how to avoid that but will look.

The production binary built with crosscompile.bash doesn't have a command-prompt window, so we should be good there.  See the -Hwindowsgui linker flag, which creates a windows rather than a command-line executable - https://github.com/getlantern/flashlight-build/blob/valencia/crosscompile.bash#L12

myleshorton:
> The production binary built with crosscompile.bash doesn't have a command-prompt window, so we should be good there. See the -Hwindowsgui linker flag, which creates a windows rather than a command-line executable - https://github.com/getlantern/flashlight-build/blob/valencia/crosscompile.bash#L12

I think it's actually the call that happens from the system tray to open the browser -- that's what flashes the command prompt.

oxtoacart:
Oh weird, even with that flag I'm seeing a command-line window shown.  I bet that's because of this - https://github.com/skratchdot/open-golang/blob/master/open/exec_windows.go

myleshorton:
Yup -- that's what I was looking for thanks. I'll open a ticket and fiddle.

oxtoacart:
@myleshorton Thanks!

oxtoacart:
@myleshorton The trayhost library uses native code to do it - https://github.com/cratonica/trayhost/blob/master/platform/windows/tray.c

myleshorton:
Oh right @oxtoacart -- so it needs to be fixed in both places. This PR in general looks fine, although now it looks like it's getting merge conflicts. You want me to merge manually, or do you want to fix the conflicts or merge manually yourself?

oxtoacart:
@myleshorton Thanks for reviewing.  I'll handle the merge.

oxtoacart:
Redoing with a different pull.

PR #76, opened 2015-02-21
use rundll32 to open default browser

fffw:
So this is the way @myleshorton found for https://github.com/getlantern/lantern/issues/2249, the simplest one.


oxtoacart:
That rundll trick is perfect!  It even works when I cross compile using ./crosscompile!

Rather than patching skratchdot/open-golang, can you please apply this to our fork at getlantern/open-golang?  Otherwise, things will get confusing in the future if we ever pull from upstream.

fffw:
Oh yes I'll fix it

oxtoacart:
LGTM!

PR #77, opened 2015-02-21
convert lantern.ico to .syso so we can embed it in windows exe

fffw:
cross compilation will bring in the .syso file automatically.

On Linux, icon  can't be embed to binary but configured through a separate Desktop Entity file, so we just leave it as is.


fffw:
This can actually resolve https://github.com/getlantern/lantern/issues/2250. For now we only care about Windows.

oxtoacart:
Very cool!

PR #79, opened 2015-02-21
Switch back to skratchdot/open-golang

oxtoacart:
skratchdot/open-golang already merged our pull request, so I've switched back to using it instead of our fork.


fffw:
LGTM.

PR #81, opened 2015-02-21
Optimizing build a little

oxtoacart:
The way I originally set up crosscompile.bash, it's running gulp every time.  I've updated it to only run gulp if the app folder has more recent changes than what's already in dist.  This way, if we didn't touch the UI, the build proceeds a lot more quickly.  I will separately check in the dist folder so that this way, anyone who just clones the repo can build immediately without gulp needing to run.

While I was in here, I also consolidated the common linker flags into a variable.


oxtoacart:
Dist folder is now checked into valencia - https://github.com/getlantern/flashlight-build/commit/960824ef590c8aeb81c70257f0ac30f59b9cd475

atavism:
Works for me.

oxtoacart:
Thanks for the quick review!

PR #80, opened 2015-02-21
Adding the ability to send multiple messages over the same /data (websocket) channel

xiam:
Hello @oxtoacart, @atavism,

Following our discussion on sending different message types over the same channel I've modified your `flashlight/ui/` and `flashlight/proxiedsites` packages, services like `proxiedsites` will need to [register themselves](https://github.com/getlantern/flashlight-build/blob/08a3a1ef30885c858db9f35b4d9d868343e758a4/src/github.com/getlantern/flashlight/proxiedsites/proxiedsites.go#L80) using `ui.Register()` and send messages using an [envelope](https://github.com/getlantern/flashlight-build/blob/08a3a1ef30885c858db9f35b4d9d868343e758a4/src/github.com/getlantern/flashlight/proxiedsites/proxiedsites.go#L66).

The client will also need to send messages in the same envelope and they can be read [almost in the same fashion](https://github.com/getlantern/flashlight-build/blob/08a3a1ef30885c858db9f35b4d9d868343e758a4/src/github.com/getlantern/flashlight/proxiedsites/proxiedsites.go#L96) that they were read before.

Lantern-UI was also updated: I added a [new service](https://github.com/getlantern/flashlight-build/blob/7fc0243c29710a6c1adc33fde552ccce426667de/src/github.com/getlantern/lantern-ui/app/js/services.js#L6) that provides a map of callbacks associated to message types and also a new [DataSource](https://github.com/getlantern/flashlight-build/blob/7fc0243c29710a6c1adc33fde552ccce426667de/src/github.com/getlantern/lantern-ui/app/js/app.js#L74) factory which is now used by [ProxiedSites](https://github.com/getlantern/flashlight-build/blob/7fc0243c29710a6c1adc33fde552ccce426667de/src/github.com/getlantern/lantern-ui/app/js/app.js#L131) to communicate with the /data channel.

This PR also contains code for closing https://github.com/getlantern/lantern/issues/2213, the UI is now receiving geolocation information and printing it to the debug output:

screenshot from 2015-02-21 08 23 09

This includes changes to both Go and Javascript code so I think it would be better is you both could help me reviewing it.

Thanks!

PR #82, opened 2015-02-21
Ability to send/receive different t messages over a single websocket channel and user's geolocation data lookup

xiam:
Hello @oxtoacart, @atavism,

My [former PR](https://github.com/getlantern/flashlight-build/pull/80) was apparently polluted with other stuff so I closed it and integrated my changes again on a fresh copy of the valencia branch.

Following our discussion on sending different message types over the same channel I've modified your `flashlight/ui/` and `flashlight/proxiedsites` packages, services like `proxiedsites` will need to [register themselves](https://github.com/getlantern/flashlight-build/blob/b3960a1698af7be13df60334e1f0d2a0041e1d9c/src/github.com/getlantern/flashlight/proxiedsites/proxiedsites.go#L80) using `ui.Register()` and send messages using an [envelope](https://github.com/getlantern/flashlight-build/blob/b3960a1698af7be13df60334e1f0d2a0041e1d9c/src/github.com/getlantern/flashlight/proxiedsites/proxiedsites.go#L66).

The client will also need to send messages in the same envelope and they can be read [almost in the same fashion](https://github.com/getlantern/flashlight-build/blob/b3960a1698af7be13df60334e1f0d2a0041e1d9c/src/github.com/getlantern/flashlight/proxiedsites/proxiedsites.go#L95) that they were read before.

Lantern-UI was also updated: I added a [new service](https://github.com/getlantern/flashlight-build/blob/ed80c984e5f8eee35e837a0d11e3f9558a26aa33/src/github.com/getlantern/lantern-ui/app/js/services.js#L6) that provides a map of callbacks associated to message types and also a new [DataSource](https://github.com/getlantern/flashlight-build/blob/b3960a1698af7be13df60334e1f0d2a0041e1d9c/src/github.com/getlantern/lantern-ui/app/js/app.js#L74) factory which is now used by [ProxiedSites](https://github.com/getlantern/flashlight-build/blob/b3960a1698af7be13df60334e1f0d2a0041e1d9c/src/github.com/getlantern/lantern-ui/app/js/app.js#L131) to communicate with the /data channel.

This PR also contains code for closing https://github.com/getlantern/lantern/issues/2213, the UI is now receiving geolocation information and printing it to the debug output:

screenshot from 2015-02-21 08 23 09

This includes changes to both Go and Javascript code so I think it would be better is you both could help me reviewing it.

Thanks!


src/github.com/getlantern/flashlight/client/balancer.go
```diff
@@ -14,9 +18,22 @@ func (client *Client) initBalancer(cfg *ClientConfig) *balancer.Balancer {
...
 		dialers = append(dialers, dialer)
+		if dialer.QOS > highestQOS {
+			highestQOSFrontedDialer = fd
+		}
+	}
+
+	if highestQOSFrontedDialer != nil {
+		// This function will configure the HTTP client the geo package is going to
+		// use.
+		geolookup.SetHTTPClient(highestQOSFrontedDialer.DirectHttpClient())
```
xiam: I don't quite like this, it does the job but it feels a bit dirty. If you have another suggestion for telling geolookup which proxy to use please elaborate, we might need that pattern for future functionality.

src/github.com/getlantern/flashlight/statserver/peer.go
```diff
@@ -35,84 +31,6 @@ type Peer struct {
 	reportedFinal   bool
 }
 
-// The City structure corresponds to the data in the GeoIP2/GeoLite2 City
```
xiam, line 38: This struct was moved to the geolookup package.

src/github.com/getlantern/flashlight/geolookup/geolookup.go
```diff
@@ -0,0 +1,94 @@
...
+
+	var err error
+
+	if lookupData != nil {
+		// We already looked up IP's information.
+		return lookupData
+	}
+
+	for {
+		if !geolookup.UsesDefaultHTTPClient() {
```
oxtoacart: So if this is false, we'll never do a geo lookup?  Why even bother starting the service?

src/github.com/getlantern/flashlight/geolookup/geolookup.go
```diff
@@ -0,0 +1,94 @@
...
+			// Will look up only if we're using a proxy.
+			lookupData, err = geolookup.LookupCity("")
```
oxtoacart: You could use waitforserver here.

src/github.com/getlantern/flashlight/client/balancer.go
oxtoacart: What don't you like about this?

src/github.com/getlantern/flashlight/client/balancer.go
```diff
@@ -14,9 +18,22 @@ func (client *Client) initBalancer(cfg *ClientConfig) *balancer.Balancer {
...
+	} else {
+		log.Debugf("No fronted dialers found, unable to look up public ip")
```
oxtoacart: Maybe error instead of debug?

src/github.com/getlantern/flashlight/ui/service.go
```diff
@@ -0,0 +1,102 @@
...
+)
+
+func (s *Service) watch() {
+	// Watch for new messages and sent them to the combined output.
+	for b := range s.Out {
+		out <- b
+	}
+}
+
+func Register(name string, helloFn func(func([]byte) error) error) (*Service, error) {
```
oxtoacart: So, as it currently stands, we have three different namespaces for messages:
1. Service Name
2. Type (on envelope)
3. URL (on UIChannel)

It seems like 1 should be enough.  What's the reason for having all 3?

src/github.com/getlantern/flashlight/ui/service.go
```diff
@@ -0,0 +1,102 @@
...
+	mu.Unlock()
+
+	return services[name], nil
+}
+
+func start() {
+	// Establish a channel to the UI for sending and receiving updates
+	defaultUIChannel = NewChannel("/data", func(write func([]byte) error) error {
+		// Sending hello messages.
+		for _, s := range services {
```
oxtoacart, line 88: Need to lock using mutex here.  If you use an RWMutex, you just need to RLock().  And, you really only need to lock, copy the list of services, unlock and then do your hello loop.

src/github.com/getlantern/flashlight/ui/service.go
```diff
@@ -0,0 +1,102 @@
...
+			// Delegating task...
+			if err := s.helloFn(write); err != nil {
+				log.Errorf("Error writing to socket: %q", err)
+			}
+		}
+		return nil
```
oxtoacart, line 103: What happens if a service is registered after a client has already connected?  Seems like you should process the helloFn at that point, right?

src/github.com/getlantern/flashlight/ui/service.go
```diff
@@ -0,0 +1,102 @@
...
+		if err != nil {
+			log.Errorf("Unable to parse JSON update from browser: %q", err)
+			continue
+		}
+
+		// Delegating response to the service that registered with the given type.
+		if services[env.Type] != nil {
+			// Pass this message and continue reading another one.
+			go func() {
+				services[env.Type].In <- b
```
oxtoacart: Because we are passing the JSON bytes rather than the unmarshalled data structure to our services, we end up having to Unmarshal twice.  It would be nice to figure out a way to only unmarshal once.  That makes the code in the services easier too since they don't have to bother with unmarshalling JSON.

src/github.com/getlantern/geolookup/geolookup.go
```diff
@@ -0,0 +1,138 @@
...
+		Type      string            `maxminddb:"type"`
+	} `maxminddb:"represented_country"`
+	Traits struct {
+		IsAnonymousProxy    bool `maxminddb:"is_anonymous_proxy"`
+		IsSatelliteProvider bool `maxminddb:"is_satellite_provider"`
+	} `maxminddb:"traits"`
+}
+
+// SetHTTPClient defines the client to use for executing HTTP requests.
+func SetHTTPClient(client *http.Client) {
```
oxtoacart: Since this can be called anytime, you probably need to protect the httpClient property with a mutex.

src/github.com/getlantern/geolookup/geolookup_test.go
```diff
@@ -0,0 +1,47 @@
...
+	}
+
+	if city.City.Names["en"] != "New York" {
+		t.Fatal("Look up failed.")
+	}
+
+}
+
+func TestUsingDefaultClient(t *testing.T) {
+	if UsesDefaultHTTPClient() == false {
```
oxtoacart: I'd recommend using github.com/getlantern/testify/assert because it makes test conditions like this easier to read.

oxtoacart: It also provides really nice output when tests fail.

src/github.com/getlantern/flashlight/geolookup/geolookup.go
oxtoacart: Also, just because it's not using the default http client doesn't mean that it's using the client that we want :)

One way to make this cleaner would be to have the geolookup.LookupCity() function accept an HttpClient. If provided, it uses that, and if not it uses the default.  That way, we can be sure that it's using the client we want when we call it, and we don't have to bother managing the httpClient variable in geolookup.

src/github.com/getlantern/flashlight/client/balancer.go
xiam: Because this SetHTTPClient() changes a variable in another package, I don't quite like working with global values that way because it's difficult to know if the current value is different from the original (if you're reading the code).

src/github.com/getlantern/flashlight/ui/service.go
xiam: Yes, but in this case service name is always equal to the message type. Regarding the URL, as far as I know, the URL is just used to name a central location for websocket clients to locate a data provider, so, we can't really use it as a namespace unless opening another connection, that leaves us with the Type/Name as the only namespace.

I realized the `Name` property of Service is actually unused and is not mean to be changed, so I'll remove it to avoid confusion.

src/github.com/getlantern/flashlight/ui/service.go
xiam, line 103: Good point! Didn't think of this case, at this time the client would not get the helloFn message since that's only sent when new connections arrive. That's very unlikely to happen but we be better close that hole now before it becomes a problem.

src/github.com/getlantern/geolookup/geolookup.go
xiam: Right!

src/github.com/getlantern/flashlight/client/balancer.go
xiam: Fixed.

src/github.com/getlantern/flashlight/ui/service.go
xiam, line 88: Great suggestion!

xiam:
I'm ready for another review!

xiam:
Just integrated the latest changes from Ox into this branch.

oxtoacart:
LGTM.

PR #83, opened 2015-02-21
Updated background image for OS X dmg

oxtoacart:
I also went ahead and vendored the rsrc tool that we use for embedding the icon in the windows executable.


atavism:
Looks great! I noticed dragging Lantern to my Applications folder overwrites the previous version of Java Lantern I had installed. Is that the desired behavior? Just confirming.

oxtoacart:
> I noticed dragging Lantern to my Applications folder overwrites the previous version of Java Lantern I had installed. Is that the desired behavior? Just confirming.

We basically punted on this by saying that we wouldn't worry about people who already had Lantern installed.  If someone who had Lantern previously installed installs the new one, there will be other problems too (e.g. what happened to my whitelist customizations?!).

That said, yes, as implemented I would expect it to replace your existing Lantern since we're using the same name for the app, and I think that's a reasonable thing.  The alternative would be calling it something else, but we couldn't agree on what to call it.

Either way, I'm really glad to be rid of the installer program on OS X, and I think Derek's background image looks great!

atavism:
> We basically punted on this by saying that we wouldn't worry about people who already had Lantern installed. If someone who had Lantern previously installed installs the new one, there will be other problems too (e.g. what happened to my whitelist customizations?!).

Oh, good point! Yeah, that's the conclusion I thought we reached, but I just wanted to confirm.

> Either way, I'm really glad to be rid of the installer program on OS X, and I think Derek's background image looks great!

+1! Looks great to me too. Merging!

oxtoacart:
Thanks for reviewing!

myleshorton:
Cool yeah and the overwriting of whatever old version of X program was there previously is standard behavior on OSX, so I think we're fine there.

PR #84, opened 2015-02-22
Suggestions for geolocation PR

oxtoacart:
Here are some suggestions.

### DIFF
--- a/src/github.com/getlantern/flashlight/client/balancer.go
+++ b/src/github.com/getlantern/flashlight/client/balancer.go
@@ -5,7 +5,6 @@ import (
 
 	"github.com/getlantern/balancer"
 	"github.com/getlantern/fronted"
-	"github.com/getlantern/geolookup"
 )
 
 func (client *Client) getBalancer() *balancer.Balancer {
@@ -14,7 +13,7 @@ func (client *Client) getBalancer() *balancer.Balancer {
 	return bal
 }
 
-func (client *Client) initBalancer(cfg *ClientConfig) *balancer.Balancer {
+func (client *Client) initBalancer(cfg *ClientConfig) (*balancer.Balancer, fronted.Dialer) {
 	dialers := make([]*balancer.Dialer, 0, len(cfg.FrontedServers)+len(cfg.ChainedServers))
 
 	log.Debugf("Adding %d domain fronted servers", len(cfg.FrontedServers))
@@ -26,14 +25,6 @@ func (client *Client) initBalancer(cfg *ClientConfig) *balancer.Balancer {
 		if dialer.QOS > highestQOS {
 			highestQOSFrontedDialer = fd
 		}
-	}
-
-	if highestQOSFrontedDialer != nil {
-		// This function will configure the HTTP client the geo package is going to
-		// use.
-		geolookup.SetHTTPClient(highestQOSFrontedDialer.DirectHttpClient())
-	} else {
-		log.Debugf("No fronted dialers found, unable to look up public ip")
 	}
 
 	log.Debugf("Adding %d chained servers", len(cfg.ChainedServers))
@@ -65,5 +56,5 @@ func (client *Client) initBalancer(cfg *ClientConfig) *balancer.Balancer {
 	client.balCh <- bal
 	client.balInitialized = true
 
-	return bal
+	return bal, highestQOSFrontedDialer
 }
--- a/src/github.com/getlantern/flashlight/client/client.go
+++ b/src/github.com/getlantern/flashlight/client/client.go
@@ -9,6 +9,7 @@ import (
 	"time"
 
 	"github.com/getlantern/balancer"
+	"github.com/getlantern/fronted"
 	"github.com/getlantern/golog"
 
 	"github.com/getlantern/flashlight/globals"
@@ -40,6 +41,7 @@ type Client struct {
 	balInitialized  bool
 	rpCh            chan *httputil.ReverseProxy
 	rpInitialized   bool
+	hqfd            fronted.Dialer
 }
 
 // ListenAndServe makes the client listen for HTTP connections
@@ -57,7 +59,9 @@ func (client *Client) ListenAndServe() error {
 
 // Configure updates the client's configuration.  Configure can be called
 // before or after ListenAndServe, and can be called multiple times.
-func (client *Client) Configure(cfg *ClientConfig) {
+// It returns the highest QOS fronted.Dialer available, or nil if none
+// available.
+func (client *Client) Configure(cfg *ClientConfig) fronted.Dialer {
 	client.cfgMutex.Lock()
 	defer client.cfgMutex.Unlock()
 
@@ -66,7 +70,7 @@ func (client *Client) Configure(cfg *ClientConfig) {
 		if reflect.DeepEqual(client.priorCfg, cfg) &&
 			reflect.DeepEqual(client.priorTrustedCAs, globals.TrustedCAs) {
 			log.Debugf("Client configuration unchanged")
-			return
+			return client.hqfd
 		} else {
 			log.Debugf("Client configuration changed")
 		}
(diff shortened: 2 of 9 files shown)


src/github.com/getlantern/geolookup/geolookup.go
```diff
@@ -96,21 +98,16 @@ type Country struct {
 	} `maxminddb:"traits"`
 }
 
-// SetHTTPClient defines the client to use for executing HTTP requests.
-func SetHTTPClient(client *http.Client) {
-	httpClient = client
+// LookupCity looks up the given IP using a geolocation service and returns a
+// City struct. If an httpClient was provided, it uses that, otherwise it uses
+// a default http.Client.
+func LookupCity(ipAddr string, httpClient *http.Client) (*City, error) {
```
oxtoacart, line 104: Instead of maintaing a package-wide client, we just allow the caller to pass it in.

src/github.com/getlantern/geolookup/geolookup_test.go
```diff
@@ -1,47 +1,31 @@
 package geolookup
 
 import (
+	"fmt"
+	"net"
 	"net/http"
 	"testing"
+
+	"github.com/getlantern/testify/assert"
```
oxtoacart, line 9: Added testify/assert for more readable test conditions.

src/github.com/getlantern/flashlight/client/balancer.go
```diff
@@ -14,7 +13,7 @@ func (client *Client) getBalancer() *balancer.Balancer {
 	return bal
 }
 
-func (client *Client) initBalancer(cfg *ClientConfig) *balancer.Balancer {
+func (client *Client) initBalancer(cfg *ClientConfig) (*balancer.Balancer, fronted.Dialer) {
```
oxtoacart, line 16: initBalancer now returns the fronted.Dialer.

src/github.com/getlantern/flashlight/client/client.go
```diff
@@ -57,7 +59,9 @@ func (client *Client) ListenAndServe() error {
 
 // Configure updates the client's configuration.  Configure can be called
 // before or after ListenAndServe, and can be called multiple times.
-func (client *Client) Configure(cfg *ClientConfig) {
+// It returns the highest QOS fronted.Dialer available, or nil if none
+// available.
+func (client *Client) Configure(cfg *ClientConfig) fronted.Dialer {
```
oxtoacart, line 64: Configure now returns the fronted.Dialer.

src/github.com/getlantern/flashlight/flashlight.go
```diff
@@ -156,7 +156,7 @@ func runClientProxy(cfg *config.Config) {
 	proxiedsites.Configure(cfg.ProxiedSites)
 
 	// launching geolookup service when in client mode.
-	go geolookup.StartService()
+	go geolookup.StartService(hqfd.DirectHttpClient())
```
oxtoacart, line 159: We pass the fronted.Dialer from client.Configure() to the geolookup service.  This way, client doesn't directly have a dependency on the geolookup service.

src/github.com/getlantern/flashlight/geolookup/geolookup.go
```diff
@@ -25,60 +24,41 @@ var (
...
-	for {
-		if !geolookup.UsesDefaultHTTPClient() {
-			// Will look up only if we're using a proxy.
-			lookupData, err = geolookup.LookupCity("")
-			if err == nil {
-				// We got what we wanted, no need to query for it again, let's exit.
-				return lookupData
-			}
+	var err error
+	for i := 0; i < maxRetries; i++ {
```
oxtoacart, line 37: I changed this logic a little.
1. Once we have a fronted dialer, we don't need to wait for any proxy to come online.  The only reason it would fail is if there's some sort of error.
2. I capped the number of retries, because as written originally this would hang indefinitely as long as it's failing.

src/github.com/getlantern/flashlight/ui/service.go
```diff
@@ -9,11 +10,22 @@ type envelopeType struct {
...
-	In      chan []byte
-	Out     chan []byte
-	helloFn func(func([]byte) error) error
+	Type       string
+	In         <-chan interface{}
+	Out        chan<- interface{}
+	in         chan interface{}
+	out        chan interface{}
+	newMessage func() interface{}
+	helloFn    helloFnType
```
oxtoacart, line 28: Several changes here.  Rather than passing around bytes, we now pass around marshalled data structures.  That way, the services themselves don't have to deal with the envelopes.

Also, I made private and public channels.  The public channels are one-way.  This is a nice pattern to prevent clients from writing to a channel that's only meant for them to read, or vice versa.

src/github.com/getlantern/flashlight/ui/service.go
```diff
@@ -25,39 +37,44 @@ var (
...
+		b, err := newEnvelope(s.Type, msg)
+		if err != nil {
+			log.Error(err.Error())
+		}
 		out <- b
 	}
 }
 
-func Register(name string, helloFn func(func([]byte) error) error) (*Service, error) {
+func Register(t string, newMessage func() interface{}, helloFn helloFnType) (*Service, error) {
```
oxtoacart, line 50: There's a new parameter, newMessage.  This is a function that's used to construct a new service-specific message into which we can unmarshal the JSON.

src/github.com/getlantern/flashlight/ui/service.go
```diff
@@ -89,14 +114,29 @@ func read() {
...
 		}
 
+		env := &envelope{}
+		err = json.Unmarshal(b, env)
+		if err != nil {
+			log.Errorf("Unable to unmarshal message of type %v: %v", envType.Type, err)
+			return
+		}
+		// Pass this message and continue reading another one.
+		services[env.Type].in <- env.Message
```
oxtoacart, line 129: Note - I'm no longer doing this in a goroutine.  The problem with spawning goroutines is that if the client started sending tons of messages, we'd end up piling up a bunch of goroutines.

A goroutine isn't necessary here anyway since the channel is buffered.

oxtoacart, line 129: To elaborate on this further, since it's an important point ...

One way to think about this is that spawning goroutines to write to a channel is essentially creating an unbounded buffer.  Go will spawn as many goroutines as it can until it runs out of memory or cpu, and if nothing is reading from the channel (or it's reading very slowly), this virtual buffer will just increase in size until it exhausts the available resources.

As a general rule, in distributed systems I prefer to back-pressure once a reasonably sized buffer has been filled. If the UI is consistently producing messages faster than the backend can consume them, the backend should back pressure to the UI rather than trying to buffer until it exhausts its own resources.  That way, the UI will slow down (or start having errors), but the backend will remain stable and able to serve other consumers (e.g. user agents that are using the proxy).  The purpose of the buffer is just to absorb temporary shocks so that if the UI experiences temporarily high activity, we can absorb that without back pressuring.

oxtoacart, line 129: A good alternative to back-pressure when the application logic allows for it is to just start dropping messages.  That pattern would look like this:

``` go
select {
    case services[env.Type].in <- env.Message:
        // there was room in the buffer, message accepted
    default:
        // no room in buffer, dropping message
}
```

xiam:
Thanks a lot Ox! I also thought on using interfaces but I could not find an easy way to do that, I like what you did.

src/github.com/getlantern/flashlight/ui/service.go
xiam, line 129: >  the backend should back pressure to the UI rather than trying to buffer until it exhausts its own resources. That way, the UI will slow down (or start having errors), but the backend will remain stable and able to serve other consumers (e.g. user agents that are using the proxy).

That's a very important point if one wants to look for stability on server side.

src/github.com/getlantern/flashlight/flashlight.go
xiam, line 159: Great addition, we'll surely need this to proxy requests to other services in the future.

PR #85, opened 2015-02-22
build pac tool as gui program on windows

fffw:


oxtoacart:
@fffw Did you already rebuild the windows binary using the new linker flag?

PR #86, opened 2015-02-22
Some bug fixes

oxtoacart:


src/github.com/getlantern/flashlight/flashlight.go
```diff
@@ -165,7 +160,8 @@ func runClientProxy(cfg *config.Config) {
 
 			proxiedsites.Configure(cfg.ProxiedSites)
 			configureStats(cfg, false)
-			client.Configure(cfg.Client)
+			hqfd = client.Configure(cfg.Client)
+			geolookup.Configure(hqfd.DirectHttpClient())
```
oxtoacart: I now continuously call geolookup.Configure so that if the hqfd changes, we get the latest.

src/github.com/getlantern/flashlight/geolookup/geolookup.go
```diff
@@ -13,58 +17,92 @@ import (
...
-	if service, err = ui.Register(messageType, nil, helloFn); err != nil {
-		return fmt.Errorf("Unable to register channel: %q", err)
-	}
+	var err error
+	service, err = ui.Register(messageType, nil, helloFn)
+	return err
+}
 
-	go read()
+func write() {
```
oxtoacart, line 71: The logic here has changed.  We now have a continuous loop that's always checking the location, waiting a random amount in between.  We do this so that if the computer's location changes, we'll know about it and update the map.

src/github.com/getlantern/flashlight/ui/service.go
```diff
@@ -22,18 +22,17 @@ var (
...
-func (s *Service) watch() {
+func (s *Service) write() {
 	// Watch for new messages and send them to the combined output.
 	for msg := range s.out {
 		b, err := newEnvelope(s.Type, msg)
 		if err != nil {
 			log.Error(err.Error())
 		}
-		out <- b
+		defaultUIChannel.Out <- b
```
oxtoacart, line 35: This old out channel didn't go anywhere, so the UI wasn't getting updates and this would eventually block once the channel was filled.  We should just write directly to the defaultUIChannel.Out.

src/github.com/getlantern/lantern-ui/app/js/app.js
```diff
@@ -75,8 +75,9 @@ var app = angular.module('app', [
     '$websocket',
     '$rootScope',
     '$interval',
+    '$window',
     'Messages',
-    function($websocket, $rootScope, $interval, Messages) {
+    function($websocket, $rootScope, $interval, $window, Messages) {
```
oxtoacart, line 80: This was missing $window, which broke the reconnection logic further down.

src/github.com/getlantern/flashlight/ui/service.go
xiam, line 35: This was a serious bug, thanks for spotting it!

src/github.com/getlantern/flashlight/geolookup/geolookup.go
```diff
@@ -13,58 +17,92 @@ import (
...
+			if !reflect.DeepEqual(location, oldLocation) {
+				log.Debugf("Location changed")
+				lastKnownLocation.Store(location)
+			}
+			// Always publish location, even if unchanged
+			service.Out <- location
+		} else {
+			log.Errorf("Unable to get current location: %v", err)
+			// When retrying after a failure, wait a different amount of time
+			retryWait := time.Duration(math.Pow(float64(retryWaitTime), float64(consecutiveFailures)))
```
xiam: Not sure if we want to use pow here, given that `retryWaitTime` is set to 100, if `consecutiveFailures` is, let's say 5 or 6 that would be a huge amount of hours, also the result could overflow fairly easily, even with small values of `consecutiveFailures`.

At this time, looks like `consecutiveFailures` is never changed (value 0), so `retryWait` is anyways always equal to 1, but in that case the above code is never user and we always end up waiting 1 nanosecond.

oxtoacart: Great catch!  Fixed.

To test this, I temporarily inserted this code on line 82:

``` go
if consecutiveFailures < 4 {
    err = fmt.Errorf("Failing intentionally")
}       
```

I verified that the wait time increased exponentially by powers of 2, that it eventually succeeded and then waited a full 30 seconds until the next call, and the went through the same retry loop.

src/github.com/getlantern/flashlight/geolookup/geolookup.go
```diff
@@ -13,58 +17,95 @@ import (
...
+			if !reflect.DeepEqual(location, oldLocation) {
+				log.Debugf("Location changed")
+				lastKnownLocation.Store(location)
+			}
+			// Always publish location, even if unchanged
+			service.Out <- location
+		} else {
+			log.Errorf("Unable to get current location: %v", err)
+			// When retrying after a failure, wait a different amount of time
+			retryWait := time.Duration(math.Pow(2, float64(consecutiveFailures)) * float64(retryWaitTime))
```
oxtoacart, line 95: @xiam Here's the new back-off logic.

xiam:
Merged Ox, thanks!

oxtoacart:
@xiam Do you know what happened to your issue-2213-2 branch?  I thought I merged it, but it doesn't look like I did.  Was there a pull request for it?

oxtoacart:
@xiam Never mind, found it.  All good!

PR #88, opened 2015-02-22
Fix rendering map in Firefox

atavism:
This fixes getlantern/lantern#2254 (comment)

The issue had to do with the custom resizable directive being applied to the map SVG element instead of its div container.


atavism:
Tested the map renders (and resizes) fine on Chrome, Firefox, and Safari.

oxtoacart:
LGTM!

PR #90, opened 2015-02-22
Packaging windows as installer

oxtoacart:
This creates an installer for windows (as opposed to a zip file).  The only problem with this implementation is that it's not localized (all English).  That's no worse than our old installer, but it's no better.


myleshorton:
Nice!! I'll give it a look.

On Sunday, February 22, 2015, Coveralls notifications@github.com wrote:

> [image: Coverage Status] https://coveralls.io/builds/1974687
> 
> Coverage remained the same at 65.17% when pulling _99113a5
> https://github.com/getlantern/flashlight-build/commit/99113a51b304223d75e2a78b6e70643a5111f11f
> on valencia-wininstall_ into _0cabfe5
> https://github.com/getlantern/flashlight-build/commit/0cabfe520f14a93b880bfb631bf2b2a5b144398b
> on valencia_.
> 
> —
> https://github.com/getlantern/flashlight-build/pull/90#issuecomment-75457197
> .

## 

## 

President
https://www.getlantern.org

oxtoacart:
@myleshorton I think this is good to go now.  It turns out that nsis automatically localizes based on the user's detected language and includes localization strings for all of its built-in prompts.  So the approach that I took is to not include any custom prompts.

As it stands, the installer does this:
1. Prompts for UAC
2. Kills any existing lantern.exe process
3. Installs to Program Files
4. Adds a shortcut to the Start Menu for All Users
5. Also adds an uninstall shortcut
6. Also adds itself to the registry so that it can be uninstalled from Add/Remove programs
7. Launches Lantern
8. Automatically dismisses itself (as long as everything was successful)

It can be built from OS X, which is where I tested the build scripts.  I've tested the resulting installer on Windows 8 and Windows XP.

oxtoacart:
@atavism Now that @myleshorton is on vacation, would you mind looking at this PR please?

atavism:
@oxtoacart Sure.

myleshorton:
I'll have a look too @atavism

lantern.nsi
```diff
@@ -0,0 +1,86 @@
...
+InstallDir $PROGRAMFILES\Lantern
+ 
+# For removing Start Menu shortcut in Windows 7
+RequestExecutionLevel admin
+    
+# start default section
+Section
+    # Stop existing Lantern if necessary
+    ${nsProcess::KillProcess} "lantern.exe" $R0
+    # Sleep for 1 second to give file lock a chance to clear
```
myleshorton: Is there a file lock?

oxtoacart: Maybe not the best wording.  The file is locked to writers by the OS while the program is running.  After killing the program, there's a short delay for that lock to clear.  That's what this refers to.

lantern.nsi
```diff
@@ -0,0 +1,86 @@
...
+    # Launch Lantern
+    ShellExecAsUser::ShellExecAsUser "" "$INSTDIR\lantern.exe"
+
+    ${nsProcess::Unload}
+SectionEnd
+ 
+# uninstaller section start
+Section "uninstall"
+    # Stop Lantern if necessary
+    ${nsProcess::KillProcess} "lantern.exe" $R0
```
myleshorton, line 95: I think you want to call ${nsProcess::Unload} here too, although it's hard to find good docs.

lantern.nsi
```diff
@@ -0,0 +1,86 @@
...
+    # Store installation folder
+    WriteRegStr HKCU "Software\Lantern" "" $INSTDIR
+
+    WriteUninstaller "$INSTDIR\uninstall.exe"
+
+    # Support uninstalling via Add/Remove programs
+    WriteRegStr HKLM "Software\Microsoft\Windows\CurrentVersion\Uninstall\Lantern" \
+                     "DisplayName" "Lantern"
+    WriteRegStr HKLM "Software\Microsoft\Windows\CurrentVersion\Uninstall\Lantern" \
+                     "UninstallString" "$\"$INSTDIR\uninstall.exe$\""
```
myleshorton, line 78: No biggie, but here you can also add an uninstall icon with:

WriteRegStr HKLM "Software\Microsoft\Windows\CurrentVersion\Uninstall\Lantern" "DisplayIcon" "path to icon or exe"

lantern.nsi
oxtoacart, line 95: Fixed, thanks.

lantern.nsi
oxtoacart, line 78: That's actually what I did at first, but having the installer and the uninstaller use the same icon didn't look quite right.  If we design an uninstall icon of our own, then this would be cool, but for now I figured the default uninstaller icon is good enough.

myleshorton, line 78: I would also do:

WriteRegStr HKLM "Software\Microsoft\Windows\CurrentVersion\Uninstall\Lantern" "Publisher" "Brave New Software Project, Inc."
WriteRegStr HKLM "Software\Microsoft\Windows\CurrentVersion\Uninstall\Lantern" "URLInfoAbout" "http://www.getlantern.org"

You could also define VERSION at the makensis command line and include:

WriteRegStr HKLM "Software\Microsoft\Windows\CurrentVersion\Uninstall\Lantern" "DisplayVersion" "${VERSION}"

myleshorton, line 78: So that would just be:

```
makensis /DVERSION=OUR_VERSION lantern.nsi || die "Unable to build installer"
```

myleshorton, line 78: Using the same icon for the uninstaller is the standard approach -- that's what Chrome, Firefox, etc do.

myleshorton, line 78: If you're worried about it for the uninstaller in Program Files, I would either put that in a subfolder or not include it at all since the standard uninstaller is from Add/Remove Programs.

myleshorton:
Couple of nits but overall looks great!

lantern.nsi
oxtoacart, line 78: About the icon - I misunderstood what you were talking about.  This registry entry for the uninstaller in Add/Remove programs is separate from the icon on the shortcut in Program Files, which is what I was talking about.  We can have our cake and eat it too.

oxtoacart:
@myleshorton I've integrated your awesome suggestions.  Should be ready to go now.

atavism:
I tried running this on Windows 8.1. It _looks_ like Lantern installs fine, but, after that, the executable never runs. I tried running just the plain executable generated from the crosscompile script, and that doesn't appear to do anything either. How much testing have we already done on Windows?

This is on a VM obviously. I have a previous version of Lantern installed on it. Could that be conflicting with the new installation somehow?

myleshorton:
Interesting @atavism. I'm also running on 8.1, and it's still working fine for me with the latest changes. Does it seem to be updating the lantern.exe in Program Files?

teamlantern:
Building makensis initially caused issues for me with a bad cert on the
scons download, but sounds like you knocked that out.

atavism:
Okay, uninstalling the previous version of Lantern and running the installer again seemed to do the trick. To verify it's actually able to override Java Lantern I uninstalled and installed our most recent Java release: this time, running our new installer again overrides the previous installation without issue. It must have been an anomaly. I'll keep playing around with it, but for now this LGTM. Merging.

oxtoacart:
Thanks for the review!

PR #92, opened 2015-02-23
Switched statserver to use websocket closes getlantern/lantern#2045

oxtoacart:


src/github.com/getlantern/flashlight/ui/service.go
```diff
@@ -0,0 +1,149 @@
...
+
+	services = make(map[string]*Service)
+)
+
+func (s *Service) write() {
+	// Watch for new messages and send them to the combined output.
+	for msg := range s.out {
+		b, err := newEnvelope(s.Type, msg)
+		if err != nil {
+			log.Error(err.Error())
```
xiam: Would not it be better to use continue instead of sending a null value to defaultUIChannel.Out? No need to send empty messages to clients.

oxtoacart: Yes!  Fixed.

xiam:
Not sure if this was introduced in this branch, but when Lantern asks for my Admin password (in OSX) and I hit Cancel, it panics and the UI just keeps showing the loading animation.

```
panic: Unable to execute /Users/reventlov/Library/Application Support/byteexec/cocoasudo: exit status 1
```

Is this a known error?

src/github.com/getlantern/lantern-ui/app/js/services.js
```diff
@@ -40,12 +88,33 @@ angular.module('app.services', [])
             $rootScope.originalList = entries;
           })
         }
-      }
+      },
+      'Stats': function(data) {
+        if (data.type != "peer") {
+          // Ignore anything that's not peers
```
xiam: This condition is not doing anything, a missing return?

oxtoacart:
@xiam That panic wasn't introduced through this branch.  Can you please log a P1 bug for it?

src/github.com/getlantern/lantern-ui/dist/js/app.js
```diff
@@ -824,11 +880,36 @@ angular.module('app.services', [])
...
+            script.src  = src;
+            script.async = true;
+            var x = document.getElementsByTagName('script')[0];
+            x.parentNode.insertBefore(script, x);
+        })();
+    };
+    $scope.loadShareScripts = function() {
+        if (!$window.twttr) {
+            // inject twitter share widget script
+            $scope.loadScript('//platform.twitter.com/widgets.js');
```
xiam, line 902: If the twitter or facebook servers are blocked for the user this would not work. Maybe we could hide social buttons until their script is available?

src/github.com/getlantern/lantern-ui/app/js/services.js
oxtoacart: Good point, fixed.  Didn't matter much right now because all messages are peers, but good to fix.

src/github.com/getlantern/lantern-ui/dist/js/app.js
oxtoacart, line 902: I'm not sure why this is showing up in the diff, as this is not something I touched as part of this branch.  I need to look into what's going on here.

atavism, line 902: The share buttons won't show up if those scripts can't be downloaded away. Besides, we should be able to proxy traffic by the time those requests are made.

atavism, line 902: Anyway*

oxtoacart, line 902: Oh, I see what happened.  This is in the dist folder, which is generated at cross compile time.  I think valencia's dist folder is just out of date.  You can safely ignore any files in this PR that are in the dist folder instead of the app folder.

xiam, line 902: Oh, cool @atavism! just making sure. 

Sure, I'll continue my review outside the dist folder then, @oxtoacart.

xiam, line 902: Maybe in a future release we could provide those scripts proxied directly from the flashlight proxy, without waiting for the system proxy settings to be set.

xiam:
Looks fine!

oxtoacart:
@xiam Thanks for the review!

PR #93, opened 2015-02-24
Issue 2232: Report Go errors to Loggly

xiam:
This PR intends to provide functionality for tunneling requests to loggly through the flashlight proxy.

See https://github.com/getlantern/lantern/issues/2232.


src/github.com/getlantern/flashlight/flashlight.go
```diff
@@ -42,6 +45,13 @@ var (
 )
 
 func init() {
+	if version == "" {
```
xiam, line 49: I moved this block to here to make sure `version` and `buildDate` are available for use in `doMain`.

src/github.com/getlantern/flashlight/flashlight.go
```diff
@@ -135,7 +141,9 @@ func runClientProxy(cfg *config.Config) {
 		ui.Show()
 	}
 
+	go logging.Configure(cfg)
```
xiam: Using a goroutine because we'll be using waitforserver and at this time the proxy is not yet ready.

src/github.com/getlantern/go-loggly/loggly.go
```diff
@@ -0,0 +1,328 @@
...
+
+const api = "https://logs-01.loggly.com/bulk/{token}"
+
+type Message map[string]interface{}
+
+var debug = Debug("loggly")
+
+var nl = []byte{'\n'}
+
+var (
```
xiam: Adding the http client we're going to use for proxying request and its associated rw mutex.

xiam:
Hello @oxtoacart, Could you help me reviewing my PR? I tested it with my loggly account and some fake error messages.

Thanks!

src/github.com/getlantern/flashlight/flashlight.go
oxtoacart: Cool.  Can you add this comment to the code please?

src/github.com/getlantern/flashlight/logging/logging.go
```diff
@@ -31,11 +39,21 @@ var (
 
 func init() {
 	lang, _ = jibber_jabber.DetectLanguage()
-	country, _ = jibber_jabber.DetectTerritory()
 	tz = time.Local.String()
 }
 
-func configureLogging() *rotator.SizeRotator {
+func Setup(version string, buildDate string) *rotator.SizeRotator {
```
oxtoacart: Why have both Setup and Configure?

src/github.com/getlantern/go-loggly/loggly.go
```diff
@@ -0,0 +1,328 @@
...
+	buffer   [][]byte
+	tags     []string
+	sync.Mutex
+}
+
+// SetHTTPClient sets the HTTP client the loggly client is going to use to
+// connect to the loggly API.
+func SetHTTPClient(newClient *http.Client) {
+	httpClientMu.Lock()
+	defaultHTTPClient = newClient
```
oxtoacart: defaultHTTPClient probably isn't the best name now that it can be changed.

oxtoacart: Also, this should probably be a property on the loggly.Client instead of a global.

oxtoacart:
@xiam Looks mostly good to me except for a few comments.  When you're ready, please also submit a PR to the original go-loggly.

src/github.com/getlantern/flashlight/logging/logging.go
xiam: `Setup` is mean to be to set `version`, `buildDate`, which are values that won't change and to initialize error writers and the loggly client. While `Configure` expects to receive a `config.Config` struct, which we can expect to change frequently and it only updates the HTTP client the client that was already created in `Setup`.

src/github.com/getlantern/flashlight/flashlight.go
xiam: Sure! done.

xiam:
Great @oxtoacart, I've pushed some fixes, could you check again?

Thanks,

src/github.com/getlantern/flashlight/logging/logging.go
```diff
@@ -60,14 +81,41 @@ func configureLogging() *rotator.SizeRotator {
 		log.Debugf("No logglyToken, not sending error logs to Loggly")
 	} else {
 		log.Debugf("Sending error logs to Loggly")
-		remoteWriter := logglyErrorWriter{loggly.New(logglyToken)}
-		errorOut = NonStopWriter(errorOut, remoteWriter)
+		logglyClient = &logglyErrorWriter{loggly.New(logglyToken)}
```
oxtoacart: Since loggly is initialized right at the beginning, but we don't get a proxied HTTP client until the call to Configure(), this theoretically means that some stuff could be sent to Loggly via an unproxied HTTP client, right?

xiam:
Hmm, Yeah, that's right @oxtoacart . I had this condition in `Flush()` which refused to continue if a client was not set, but I removed it thinking on that it won't be useful if we created a PR to the original client (and forgot the original purpose). I think the fastest solution would be adding that condition again (and keeping our fork, that way the messages will be tried again until a client is available, no more changes needed), but we could also drop logs when a proxy is not available, or make `writeToLoggly` wait until we have a client (that last one would block error logging until a client appears since buffering is done within the loggly package).

Do you see another option?

oxtoacart:
LGTM!

PR #94, opened 2015-02-24
Display current version of lantern in the UI

atavism:
This PR includes a new service for sending Lantern's current settings to the UI. Right now, the only info being sent is the version number and build date, but we can modify this next to include user-specific settings such as whether or not Lantern should run on system start, if usage should be anonymously reported, etc.


oxtoacart:
LGTM!  This multiplexed services over websocket thing is working out nicely :)

PR #96, opened 2015-02-24
getlantern/lantern#2232 Code review fixes

oxtoacart:
@xiam Suggested updates for your logging implementation


crosscompile.bash
```diff
@@ -8,7 +8,7 @@ VERSION="`git describe --abbrev=0 --tags --exact-match || git rev-parse --short
 BUILD_DATE="`date -u +%Y%m%d%.%H%M%S`"
 export VERSION_STRING="$VERSION ($BUILD_DATE)"
 LOGGLY_TOKEN="469973d5-6eaf-445a-be71-cf27141316a1"
-LDFLAGS="-w -X main.version $VERSION -X main.buildDate $BUILD_DATE -X main.logglyToken LOGGLY_TOKEN"
+LDFLAGS="-w -X main.version $VERSION -X main.buildDate $BUILD_DATE -X github.com/getlantern/flashlight/logging.logglyToken $LOGGLY_TOKEN"
```
oxtoacart, line 11: This never worked, so nothing ever actually got logged to Loggly.  I fixed it to embed the token correctly and verified that we're getting data to Loggly now.

src/github.com/getlantern/flashlight/logging/logging.go
```diff
@@ -23,40 +23,27 @@ import (
...
 
-	// A wrapper of the loggly client for the current session.
-	logglyClient *logglyErrorWriter
+	errorOut io.Writer
+	debugOut io.Writer
+
+	lastAddr string
 )
 
 func init() {
```
oxtoacart, line 46: All the basic logging stuff can be done without any external data, so I'm doing it in init() now.

src/github.com/getlantern/flashlight/logging/logging.go
```diff
@@ -67,108 +54,122 @@ func Setup(version string, buildDate string) *rotator.SizeRotator {
...
+			versionToLoggly: fmt.Sprintf("%v (%v)", version, buildDate),
+			client:          loggly.New(logglyToken),
+		}
+		logglyWriter.client.SetHTTPClient(client)
+		addLoggly(logglyWriter)
+	}
+}
 
-	return nil
+func Close() error {
```
oxtoacart, line 102: Instead of returning the logfile so that main can close it, we just expose a Close() function which main can use instead.

src/github.com/getlantern/flashlight/logging/logging.go
```diff
@@ -67,108 +54,122 @@ func Setup(version string, buildDate string) *rotator.SizeRotator {
...
+
+		lang, _ := jibber_jabber.DetectLanguage()
+		logglyWriter := &logglyErrorWriter{
+			lang:            lang,
+			tz:              time.Local.String(),
+			versionToLoggly: fmt.Sprintf("%v (%v)", version, buildDate),
+			client:          loggly.New(logglyToken),
+		}
+		logglyWriter.client.SetHTTPClient(client)
+		addLoggly(logglyWriter)
```
oxtoacart: Configure() now builds a logglyWriter on the fly every time that the cfg.Addr has changed.  logglyWriter has all the state that it needs (e.g. country, tz, etc.) and none of that is global anymore.

src/github.com/getlantern/flashlight/logging/logging.go
```diff
@@ -67,108 +54,122 @@ func Setup(version string, buildDate string) *rotator.SizeRotator {
...
-	l *loggly.Client
+func addLoggly(logglyWriter io.Writer) {
+	golog.SetOutputs(io.MultiWriter(errorOut, logglyWriter), debugOut)
 }
 
-func (w logglyErrorWriter) Write(b []byte) (int, error) {
-	return writeToLoggly(w.l, "ERROR", string(b))
+func removeLoggly() {
+	golog.SetOutputs(errorOut, debugOut)
 }
```
oxtoacart, line 154: addLoggly and removeLoggly allow adding a logglyWriter to the base errorOut Writer.  This way, we can add remove Loggly whenever we're ready (e.g. once Configure() has been called) and we can change the Loggly settings simply by creating a new logglyWriter and adding that.

src/github.com/getlantern/flashlight/logging/logging.go
```diff
@@ -67,108 +54,122 @@ func Setup(version string, buildDate string) *rotator.SizeRotator {
...
-	err := l.Send(m)
+
+	err := w.client.Send(m)
 	if err != nil {
 		return 0, err
 	}
-	return len(msg), nil
-}
-
-type nonStopWriter struct {
```
oxtoacart, line 157: I took out nonStopWriter because I noticed that my errors weren't making it to the console.  I replaced it with io.MultiWriter, which has the downside that it stops as soon as any of the writers had an error, but the upside is that it works and my output makes it to the console.

crosscompile.bash
xiam, line 11: Thanks Ox! that's a testing-only token, right?

oxtoacart, line 11: Nope, that's the production token.

xiam, line 11: Hmm... that's a thing. Since that's the only thing one needs to post to loggly I was going to suggest we remove it from the public view and add it to too-many-secrets but then I realized it will be embedded into the binary anyway, so I guess we may as well leave it public for now.

oxtoacart, line 11: Yeah, by definition it has to be public because our clients will use it to post.

oxtoacart:
@xiam How does it look?

oxtoacart:
@xiam I just thought of one more improvement.  Will check in shortly.

oxtoacart:
@xiam Okay, change is made.

src/github.com/getlantern/flashlight/logging/logging.go
```diff
@@ -67,108 +54,132 @@ func Setup(version string, buildDate string) *rotator.SizeRotator {
...
 	}
 
-	logglyClient.l.SetHTTPClient(cli)
+	log.Debugf("Sending error logs to Loggly via proxy at %v", cfg.Addr)
 
-	return nil
+	lang, _ := jibber_jabber.DetectLanguage()
+	logglyWriter := &logglyErrorWriter{
+		lang:            lang,
+		tz:              time.Now().Format("MST"),
```
oxtoacart, line 140: The way we were getting the time zone before just printed "local".  Now this prints a timezone string like "CST".

src/github.com/getlantern/flashlight/logging/logging.go
```diff
@@ -67,108 +54,132 @@ func Setup(version string, buildDate string) *rotator.SizeRotator {
...
+	cfgMutex.Lock()
+	if cfg.Addr == lastAddr {
+		cfgMutex.Unlock()
+		log.Debug("Logging configuration unchanged")
+		return
 	}
 
+	// Using a goroutine because we'll be using waitforserver and at this time
+	// the proxy is not yet ready.
+	go func() {
```
oxtoacart, line 95: I moved the goroutine into here so that callers don't need to be aware of the nuance.

xiam:
Thanks @oxtoacart, I'm into it.

src/github.com/getlantern/flashlight/logging/logging.go
xiam, line 140: We're forcing UTC in line 109:

```
+       return fmt.Fprintf(w, "%s - ", time.Now().In(time.UTC).Format(logTimestampFormat))
```

Should not this timezone match the actual date we're logging?

oxtoacart, line 140: Great question.  We actually use the timezone to identify the user's location in case geolocation doesn't work, so we actually want the user's timezone, not the reported time.

When you look in Loggly, you can see that the timestamp is displayed with its timezone, so there should be little confusion.

I am going to rename timeZone to userTimeZone just to make it clearer.

screen shot 2015-02-25 at 11 11 12 am

oxtoacart, line 140: Actually, our old logs already use the name "timeZone", so I'm going to leave it alone.

xiam, line 140: OK! cool.

src/github.com/getlantern/systray/systray_windows.go
```diff
@@ -4,15 +4,17 @@ import (
 	"fmt"
```
xiam, line 4: I guess this systray_windows made it to here by accident, right?

oxtoacart, line 4: Yup.

xiam:
@oxtoacart Looks good to me and I've just checked that the systray_windows.go is the same one that is on the valencia branch (after the last merge), so it won't be a problem.

xiam:
Merged, let me check my branch again. Thank you @oxtoacart!

oxtoacart:
@xiam Thanks!

src/github.com/getlantern/flashlight/logging/logging.go
fffw: @oxtoacart lantern.log never gets written on Windows 7 as we use '-Hwindowsgui' ldflag. If I remove the flag or revert back to `NonStopWriter` everything works fine.
I suspect that's because writing to `os.Stderr` always fails with that flag, hence `io.MultiWriter` will also fail.
What's your environment when errors not get to console?

oxtoacart: OS X

PR #105, opened 2015-02-27
add google domains to genconfig

fffw:
This resolves https://github.com/getlantern/lantern/issues/2269

Lots of masquerade ip addresses changed, as I connected to a VPN at Singapore when running `genconfig`. If that's not what we want, please reset latest commit and generate it again.


oxtoacart:
@fffw The masquerade config changes a lot every time we run it, so don't worry about that.

myleshorton:
Not sure why yet @fffw, but those new sites aren't making into the served pac file.

myleshorton:
OK, it looks like the centralized config was overriding the local one. I just updated the centralized one, but it looks like that's not getting picked up because of #2274

PR #109, opened 2015-03-01
Just adds a simple run script closes getlantern/lantern#2277

myleshorton:


run.bash
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+
+source setenv.bash
+pushd src/github.com/getlantern/flashlight
+go build
```
oxtoacart: Probably safer to say `go build && ./flashlight` so that it doesn't run flashlight when the build fails.

PR #110, opened 2015-03-01
Including more ip prefixes as fallbacks, closes getlantern/lantern#1739

oxtoacart:


src/github.com/getlantern/peerscanner/web.go
```diff
@@ -149,12 +149,29 @@ func clientIpFor(req *http.Request, name string) string {
 	ip := strings.TrimSpace(ips[0])
 	// TODO: need a more robust way to determine when a non-fallback host looks
 	// like a fallback.
-	if !isFallback(name) && strings.HasPrefix(ip, "128.199") {
```
oxtoacart, line 152: When I merged the changes from devel, I pulled in this very narrow definition of what is a fallback.  I've expanded it to include ips that are in use in the CloudFlare Amsterdam data center.

src/github.com/getlantern/peerscanner/group.go
```diff
@@ -39,13 +39,13 @@ func (g *group) deregister(h *host) {
 
 	log.Debugf("Deregistering from %v: %v", g.subdomain, h)
 
+	g.existing = nil
```
oxtoacart, line 42: @myleshorton I also realized that in theory, if DestroyRecord indicates a failure but actually did destroy the record, we would end up with a problem when re-registering this group since we'd still see isProxying = true and thus skip the registration.  So, I've moved the clearing of these variables before the error/return.

src/github.com/getlantern/peerscanner/host.go
```diff
@@ -346,14 +346,14 @@ func (h *host) deregisterHost() {
 }
 
 func (h *host) doDeregisterHost() error {
+	h.record = nil
```
oxtoacart, line 349: Same thing that's true for deregistering a group holds true for the host as well.

src/github.com/getlantern/peerscanner/group.go
oxtoacart, line 42: BTW, this isn't entirely academic because when I saw the 409s on fallbacks, I saw errors in the log indicating that requests to CloudFlare to deregister from rotations were failing with 524 Origin Time-out.  If the request was actually received and processed, and timed out on the response phase, that would explain the 409's (which were caused by the nl.fallbacks.getiantem.org rotation being empty).

myleshorton, line 42: nice makes sense.

src/github.com/getlantern/peerscanner/web.go
```diff
@@ -149,12 +149,29 @@ func clientIpFor(req *http.Request, name string) string {
...
+	}
+
+	return false
+}
+
+var fallbackIPPrefixes = []string{
+	"128.199",
+	"178.62",
+	"188.166",
+}
```
myleshorton, line 177: nice.

myleshorton:
LGTM

oxtoacart:
@myleshorton Thanks for the review!

myleshorton:
my pleasure!

On Mon, Mar 2, 2015 at 9:51 AM, oxtoacart notifications@github.com wrote:

> @myleshorton https://github.com/myleshorton Thanks for the review!
> 
> —
> https://github.com/getlantern/flashlight-build/pull/110#issuecomment-76676506
> .

## 

## 

President
https://www.getlantern.org

src/github.com/getlantern/peerscanner/web.go
aranhoide, line 177: Dumb question: are we sure that no peer could have an IP matching this?  Or is this just some extra sanity check?  

If we want to be positive about whether a server is a fallback, we could have it prove so upon registering, e.g. by providing a secret token that we deploy to our servers.

Something would need to be done for peerscanner startup to decide what to do in the gap until servers first register.  One option is that on startup peerscanner just asks all servers with fallback-like names to register, and it doesn't go into the main loop until all have done so or there's some timeout.

Another option is that we just treat servers with fallback-like names as fallbacks until they first register.

To determine whether any of this is warranted, I'd need to refresh myself on the purpose(s) for which peerscanner is using the isFallback information.  Was that only so we don't deregister their non-round-robin entries when checks fail for them?

oxtoacart, line 177: > Dumb question: are we sure that no peer could have an IP matching this?  Or is this just some extra sanity check? 

Great question actually.  This is not a certain check, hence my TODO comment in the code.

> If we want to be positive about whether a server is a fallback, we could have it prove so upon registering, e.g. by providing a secret token that we deploy to our servers.

Yes, that would be one way to make it more robust.

> Something would need to be done for peerscanner startup to decide what to do in the gap until servers first register.  One option is that on startup peerscanner just asks all servers with fallback-like names to register, and it doesn't go into the main loop until all have done so or there's some timeout.

We're already in a state where servers are registered.  On startup, peerscanner pulls all that information in from CloudFlare and as long as the already registered servers pass the test, they remain in rotation.

> Another option is that we just treat servers with fallback-like names as fallbacks until they first register.

We already treat servers with fallback-like names as fallbacks, except if they don't have a fallback IP.

> To determine whether any of this is warranted, I'd need to refresh myself on the purpose(s) for which peerscanner is using the isFallback information.  Was that only so we don't deregister their non-round-robin entries when checks fail for them?

Mainly this is to prevent a situation that we saw occasionally in which a peer would register via using a call proxied via a fallback.

aranhoide, line 177: > >  Something would need to be done for peerscanner startup to decide what to do in the gap until servers first register. One option is that on startup peerscanner just asks all servers with fallback-like names to register, and it  doesn't go into the main loop until all have done so or there's some timeout.
> 
> We're already in a state where servers are registered. On startup, peerscanner pulls all that information in from CloudFlare and as long as the already registered servers pass the test, they remain in rotation.

I'm not sure I said it right..  I was explaining a problem with my proposal that servers provide this token upon registering: that on peerscanner startup, there is a gap of time when peerscanner can learn about server names and IPs, but not about whether these are real fallbacks (unless we find some way to store this info somewhere convenient enough).

> Mainly this is to prevent a situation that we saw occasionally in which a peer would register via using a call proxied via a fallback.

I don't understand.  In this scenario I'd imagine the server name is more reliable than the IP? How would the fallback in the middle get to mess with the name?

oxtoacart, line 177: > I don't understand. In this scenario I'd imagine the server name is more reliable than the IP? How would the fallback in the middle get to mess with the name?

Sorry, there are two scenarios:
1. Peer registering through fallback and accidentally getting its IP
2. Malicious peer registering itself with a non-fallback IP but a fallback name (and thus getting into the fallbacks rotation)

I'm attempting to protect against both, but the protection for no. 2 is fairly week since someone could set up a peer that has a fallback-like IP (for example by putting it on DO).

aranhoide, line 177: Yeah, I don't think IP matching can work for no.2.

Also, it's a maintenance issue.  When adding fallbacks (esp. if they're in different datacenters) we must check whether their IPs are included in the IP ranges hardcoded into peerscanner.  We can only do this after they've launched.  This makes adding flashlight servers a multi-step process.

Then again, I wouldn't mess much further with this until (i) we find this hurts much, or (ii) we've decided on peers.  For example, if we stop supporting give mode peers then we won't need to tell peers from fallbacks at all.

PR #111, opened 2015-03-03
Fix log rotation (required forking rotator)

oxtoacart:


src/github.com/getlantern/rotator/size_rotator.go
```diff
@@ -40,24 +40,25 @@ func (r *SizeRotator) Write(bytes []byte) (n int, err error) {
...
-				if err != nil {
-					return 0, err
-				}
-				if r.file != nil {
-					// reset file reference
-					r.file.Close()
-					r.file = nil
-				}
-				break
+		r.file.Close()
```
myleshorton: Looks like we still have to check for nil here.

aranhoide: ~~What if `r.file == nil`?  Since you've checked for that some lines up, I guess that's a possible condition?~~ whoops, @myleshorton beat me to it.

src/github.com/getlantern/rotator/size_rotator.go
```diff
@@ -40,24 +40,25 @@ func (r *SizeRotator) Write(bytes []byte) (n int, err error) {
...
+		err := os.Remove(dpath)
+		if err != nil && !os.IsNotExist(err) {
+			return 0, fmt.Errorf("Unable to delete oldest file: %v", err)
+		}
+
+		// Rename existing files
+		for i := r.MaxRotation - 1; i >= 0; i-- {
+			opath := r.path
+			if i != 0 {
+				opath = opath + "." + strconv.Itoa(i)
```
aranhoide, line 58: Nit: a `func rotationPath(basePath string, number int)` might make things marginally clearer.  We could also add a comment there for the special case where number == 0, because within that context it'd be less distracting.

aranhoide:
Other than the failure to check for a `nil` file when size exceeded, LGTM.

PR #112, opened 2015-03-03
Removed peers

oxtoacart:


oxtoacart:
@myleshorton FYI - here's the PR for removing peers

PR #114, opened 2015-03-04
detect and proxy automatically

fffw:
It's just an experiment to overcome drawbacks of fixed whitelist. Lacks of doc, test and contains lots of bugs but does work for me very well except it always time out accessing YouTube on Chrome (don't know why).

Basically, it dial directly and read from remote first with a timeout (1s), if failed, resend buffered bytes through proxy to see if that works. It also add detoured site to a in memory whitelist, but not persisted yet.

`TRACE=detour ./lantern_darwin_amd64` to show how it runs.

Seen lots of EOFs through trace log, not knowing what cause it.

Need effort to make it mature, but may be worth to have a try.


src/github.com/getlantern/detour/detour.go
```diff
@@ -0,0 +1,288 @@
...
+// Write() implements the function from net.Conn
+func (dc *detourConn) Write(b []byte) (n int, err error) {
+	if dc.state == stateInitial {
+		if n, err = dc.writeToBuffer(b); err != nil {
+			return n, fmt.Errorf("Unable to write to local buffer: %s", err)
+		}
+	}
+	if n, err = dc.getConn().Write(b); err != nil {
+		log.Debugf("Write %d bytes to %s %s failed: %s", len(b), dc.addr, dc.stateDesc(), err)
+		if dc.state != stateDetour {
```
oxtoacart: Needs mutex.

oxtoacart: Your you can use the sync/atomic package.

src/github.com/getlantern/detour/detour.go
```diff
@@ -0,0 +1,288 @@
...
+		}
+	}
+	atomic.CompareAndSwapUint32(&dc.state, stateInitial, stateDirect)
+	log.Tracef("Read %d bytes from %s directly, set state to %s", n, dc.addr, dc.stateDesc())
+	return n, nil
+}
+
+// Write() implements the function from net.Conn
+func (dc *detourConn) Write(b []byte) (n int, err error) {
+	if dc.state == stateInitial {
```
oxtoacart: Need mutex.

oxtoacart: BTW, a handy way to check if you have possible concurrency issues is to compile/run with the `-race` flag, which enables race detection.

src/github.com/getlantern/detour/detour.go
oxtoacart: Oh, I just noticed that you're using sync/atomic to set it.  You should also use sync/atomic to load it.

src/github.com/getlantern/detour/detour.go
```diff
@@ -0,0 +1,288 @@
...
+}
+
+// Read() implements the function from net.Conn
+func (dc *detourConn) Read(b []byte) (n int, err error) {
+	conn := dc.getConn()
+	if dc.state != stateInitial {
+		if n, err = conn.Read(b); err != nil {
+			log.Debugf("Read from %s %s failed: %s", dc.addr, dc.stateDesc(), err)
+			if dc.state != stateDetour {
+				log.Debugf("Add %s to white list temporarily", dc.addr)
```
oxtoacart: Should this be temporary or permanent?

oxtoacart:
Very cool!  It's neat how you encapsulated this in a net.Conn.

One thing I wonder about though is that some forms of blocking aren't done at the TCP level.  For example, in Iran, when a site is blocked, they simply redirect you to a different page.  At the TCP level, everything looks fine, but at the HTTP level you're getting a redirect to a specific page (which is pretty easy to look for).

src/github.com/getlantern/detour/detour.go
fffw: Still need to change the logic here or many sites will end up into whitelist.

fffw:
I think we can combine detection at HTTP and TCP layer with fixed whitelist (well known and lantern related sites) to make the whole suit. It would be great to have a tool that just works in any corner of earth.

It has a side effect of accelerating some sites which are very slow to visit directly, needs tune to balance.

BTW, is `detour` a good name? I'm not confident on naming.

src/github.com/getlantern/detour/detour.go
fffw: Thanks, I totally missed that.

src/github.com/getlantern/golog/golog.go
```diff
@@ -79,7 +80,17 @@ func LoggerFor(prefix string) Logger {
 	l := &logger{
 		prefix: prefix + ": ",
 	}
-	l.traceOn, _ = strconv.ParseBool(os.Getenv("TRACE"))
+	trace := os.Getenv("TRACE")
```
fffw: I made a change here so instead of print all traces we can trace  selected packages
`TRACE=flashlight, enproxy, detour ...`

src/github.com/getlantern/flashlight/pac.go
```diff
@@ -41,16 +44,33 @@ func setUpPacTool() error {
 	return nil
 }
 
+func setProxyAddr(addr string) {
```
fffw: PAC file is fixed now, so move it here and remove all those code to generate PAC file.

src/github.com/getlantern/flashlight/proxiedsites/proxiedsites.go
```diff
@@ -25,10 +26,13 @@ var (
...
-func Configure(cfg *proxiedsites.Config, proxyAddr string) {
-	delta := proxiedsites.Configure(cfg, proxyAddr)
+func Configure(cfg *proxiedsites.Config) {
+	delta := proxiedsites.Configure(cfg)
 	startMutex.Lock()
 
+	if delta != nil {
+		updateDetour(delta)
+	}
 	if service == nil {
```
fffw: We may remove the chunks of code to maintain custom whitelist at all, but I choose not to touch them now.

src/github.com/getlantern/golog/golog.go
oxtoacart: That's very cool!

src/github.com/getlantern/flashlight/pac.go
```diff
@@ -9,11 +10,13 @@ import (
 	"github.com/getlantern/filepersist"
 	"github.com/getlantern/pac"
 
-	"github.com/getlantern/flashlight/proxiedsites"
+	"github.com/getlantern/flashlight/ui"
```
oxtoacart: This is a strange dependency.  I like that you made detour its own library, which makes it useful to other people.  However, if it has dependencies on flashlight, it becomes less useful.  I would invert the dependency so that the caller of detour passes it a callback (or something similar) which detour can use to update the PAC file stuff.

src/github.com/getlantern/flashlight/proxiedsites/proxiedsites.go
```diff
@@ -52,6 +56,24 @@ func Configure(cfg *proxiedsites.Config, proxyAddr string) {
 	startMutex.Unlock()
 }
 
+func updateDetour(delta *proxiedsites.Delta) {
+	// TODO: subscribe changes of geolookup and set country accordingly
```
fffw: Significant refactor is required to eliminate globals and unnecessary dependencies. I'll log a new issue for it.

src/github.com/getlantern/flashlight/proxiedsites/proxiedsites.go
```diff
@@ -52,6 +56,24 @@ func Configure(cfg *proxiedsites.Config, proxyAddr string) {
...
+	// safe to hardcode here as IR has all detection rules
+	detour.SetCountry("IR")
```
oxtoacart: Rather than encoding country into detour itself, it may make sense to provide a menu of detection rules from which the caller can pick and choose when initializing detour.  If the caller wants to manage it on a per-country basis, it can.

oxtoacart: Also, since there's just one detector right now, it might be easiest just to get rid of the country thing completely and only revisit this if and when we find the need to have different detection algorithms.

src/github.com/getlantern/detour/detect.go
```diff
@@ -0,0 +1,65 @@
...
+	// for the facts behind detection rules for Iran
+	detectors["IR"] = &Detector{
+		CheckConn: func(c net.Conn) bool {
+			if ra := c.RemoteAddr(); ra != nil {
+				return ra.String() == iranRedirectAddr
+			}
+			return false
+		},
+		CheckContent: func(b []byte) bool {
+			return bytes.HasPrefix(b, http403) && bytes.Contains(b, iranIFrame)
```
oxtoacart: This won't work for encrypted (HTTPS) pages.

src/github.com/getlantern/flashlight/proxiedsites/proxiedsites.go
fffw: Two detector actually. One for Iran and one for all other countries. Hardcode IR will add small performance penalty for other countries.

oxtoacart: But you're only using the one for Iran right now, right?

src/github.com/getlantern/flashlight/pac.go
fffw: Sorry, but it's `pac.go` depends on a sub-package here. detour doesn't have any dependency on flashlight.

oxtoacart: Ahh, awesome!

src/github.com/getlantern/flashlight/pac.go
oxtoacart: So this is still an outstanding item?

src/github.com/getlantern/flashlight/proxiedsites/proxiedsites.go
fffw: The detection rule itself is rather simple and it's easy to change detect.go.  Addition of rules may requires change of detour's logic. I think it just works now.

src/github.com/getlantern/detour/detect.go
fffw: Iran doesn't hijack content for https connection at all, see [my simple test result](https://github.com/getlantern/lantern/issues/2099#issuecomment-78015418)

In general, are there any way to hijack content in a https session?

oxtoacart: > Iran doesn't hijack content for https connection at all

Cool.  That might be worth a comment here.

> In general, are there any way to hijack content in a https session?

Yes.  Iran in particular is pushing people to use their "National Browser" which will trust stuff that's been MITM'ed by the government - http://www.iranhumanrights.org/2014/11/internet-reportthe-national-browser/

oxtoacart: Also, I think I heard someone in Valencia mention that Iran has started redirecting to sites other than the standard redirect (even sometimes real sites).  Detecting that may be hard, but that's probably okay, since even if detour only detects blocking sometimes, that's better than never.

src/github.com/getlantern/detour/LICENSE
```diff
@@ -0,0 +1,202 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "{}"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright {yyyy} {name of copyright owner}
```
oxtoacart: 2015 Brave New Software Project, Inc.

src/github.com/getlantern/flashlight/pac.go
```diff
@@ -9,11 +10,13 @@ import (
...
 )
 
 var (
 	isPacOn = int32(0)
```
oxtoacart: This is never read, only set.

src/github.com/getlantern/detour/detour.go
```diff
@@ -0,0 +1,353 @@
...
+func Dialer(d dialFunc) dialFunc {
+	return func(network, addr string) (conn net.Conn, err error) {
+		dc := &Conn{dialDetour: d, network: network, addr: addr}
+		if !whitelisted(addr) {
+			detector := blockDetector.Load().(*Detector)
+			dc.setState(stateInitial)
+			// always try direct connection first
+			dc.conn, err = net.DialTimeout(network, addr, TimeoutToDetour)
+			if err == nil {
+				if !detector.CheckConn(dc.conn) {
```
oxtoacart: Might want to rename CheckConn to something more meaningful like IsHijacked.

src/github.com/getlantern/detour/detour.go
```diff
@@ -0,0 +1,353 @@
...
+					log.Tracef("Dial %s to %s succeeded", dc.stateDesc(), addr)
+					return dc, nil
+				}
+				log.Debugf("Dial %s to %s, dns hijacked, try detour", dc.stateDesc(), addr)
+			} else if detector.CheckError(err) {
```
oxtoacart: May want to use more meaningful name than CheckError, for example TamperingSuspected.

src/github.com/getlantern/detour/detour.go
```diff
@@ -0,0 +1,353 @@
...
+
+	detector := blockDetector.Load().(*Detector)
+	if err != nil {
+		log.Debugf("Error while read from %s %s: %s", dc.addr, dc.stateDesc(), err)
+		if detector.CheckError(err) {
+			return dc.detour(b)
+		}
+		return
+	}
+	if detector.CheckContent(b) {
```
oxtoacart: If the content that CheckContent is looking for is split across 2 reads, this won't work.

src/github.com/getlantern/detour/detour.go
```diff
@@ -0,0 +1,353 @@
...
+		log.Tracef("Read %d bytes from %s %s, content is hijacked, detour", n, dc.addr, dc.stateDesc())
+		return dc.detour(b)
+	}
+	log.Tracef("Read %d bytes from %s %s, set state to direct", n, dc.addr, dc.stateDesc())
+	dc.setState(stateDirect)
+	return
+}
+
+func (dc *Conn) followUpRead(b []byte) (n int, err error) {
```
oxtoacart: This would benefit from a comment.

src/github.com/getlantern/detour/detour.go
```diff
@@ -0,0 +1,353 @@
...
+		case dc.inState(stateDirect) && detector.CheckError(err):
+			log.Tracef("%s still blocked, add to whitelist so will try detour next time", dc.addr)
+			addToWl(dc.addr, false)
+		case dc.inState(stateDetour) && wlTemporarily(dc.addr):
+			log.Tracef("Detoured route is not reliable for %s, not whitelist it", dc.addr)
+			removeFromWl(dc.addr)
+		}
+		return
+	}
+	if dc.inState(stateDirect) && detector.CheckContent(b) {
```
oxtoacart: Same comment as above, if the content is split between two reads, this won't work.

src/github.com/getlantern/detour/detour.go
```diff
@@ -0,0 +1,353 @@
...
+	log.Tracef("Read %d bytes from %s %s", n, dc.addr, dc.stateDesc())
+	return
+}
+
+func (dc *Conn) detour(b []byte) (n int, err error) {
+	if err = dc.setupDetour(); err != nil {
+		log.Errorf("Error while setup detour: %s", err)
+		return
+	}
+	if _, err = dc.resend(); err != nil {
```
oxtoacart: This seems dangerous.  Can this result in double submits?

src/github.com/getlantern/flashlight/pac.go
fffw: Sorry I mean PAC file no longer includes proxied sites, so I move it here. Just a hint.

src/github.com/getlantern/detour/detect.go
fffw: Yeah we have to keep pace with the censors, once we have concrete example.

src/github.com/getlantern/detour/LICENSE
fffw: Yeah while making changes I found a few repos without LICENSE file, add them as well? We might also keep the year up to date.

src/github.com/getlantern/flashlight/pac.go
fffw: It is checked at L77

src/github.com/getlantern/detour/detour.go
fffw: The content is more likely to be encapsulated in one IP packet, so I assume it will be returned in single read. Checking consecutive reads might be too heavy.

I'll comment it. Might an more accurate method name help? I'm not good at it though.

src/github.com/getlantern/detour/detour.go
fffw: Oh I never noticed this critical point. Thank you so much @oxtoacart .

ssl handshake will not succeed if no reads right? So this affects some plain sites which is not blocked but with large latency, if the very first operation in a connection to the site is non-idempotent (i.e. POST). I don't expect it to be many but it's harmful indeed. The influence to web server is also unclear when exactly same submission comes from total different IP.

OTOH, detour without application noticed is very handy. Just leave it and see what happens?

oxtoacart: Although it's not very likely, the worst case scenario here is a double form submission for a financial transaction, which would be very bad.  I think the safest thing to do is take it out.

At some point, I think we're going to want to have some detour-like logic that's aware of the HTTP protocol, to handle stuff like this and other potential things that are easier to detect at the HTTP level.

fffw: I can't imagine a financial transaction without HTTPS protection, but for safety and standard compatibility, as per section 8.1.4 of [rfc2068](https://www.ietf.org/rfc/rfc2068.txt), we MUST check and only resend those idempotent methods.

HTTP parsing might be time consuming, but fortunately, HTTP method is always at the very beginning of a request, so we can check it through byte match.

src/github.com/getlantern/flashlight/pac.go
myleshorton: Wait sorry what does the PAC file include if not proxied sites @fffw?

fffw: Oh I mean it just proxy all traffic except localhost, so it no longer depends on proxied sites list. Sorry the confusion caused.

myleshorton: Wait so what impact does this have on unblocked sites? The whole point of the PAC file is to make sure we don't slow down unblocked sites. Even if we go directly from Lantern, how much impact does that have versus going directly from the browser without Lantern for unblocked sites?

fffw: <10ms in average. See simple benchmark [here]https://github.com/getlantern/lantern/issues/2099#issuecomment-78971159

If we choose auto-detection, we have to pass all traffic through Lantern.

myleshorton: OK I don't that aspect, as we bypass the many optimizations the browser does to speed up site loading, but let's go with it and see what users think.

oxtoacart: Using an extension could help with this.

oxtoacart: In fact, an interesting project would be an extension-based version of detour.  In addition to the potential performance benefit, it would also allow for doing detection on HTTPS sites.

myleshorton: Yes! That sounds right to me.

On Saturday, March 14, 2015, oxtoacart <notifications@github.com> wrote:

> In src/github.com/getlantern/flashlight/pac.go
> <https://github.com/getlantern/flashlight-build/pull/114#discussion_r26438559>
> :
>
> > @@ -41,16 +44,33 @@ func setUpPacTool() error {
> >  	return nil
> >  }
> >
> > +func setProxyAddr(addr string) {
>
> In fact, an interesting project would be an extension-based version of
> detour. In addition to the potential performance benefit, it would also
> allow for doing detection on HTTPS sites.
>
> —
> <https://github.com/getlantern/flashlight-build/pull/114/files#r26438559>.
>


-- 
--
President
https://www.getlantern.org

fffw: Hmm, from http://www.aosabook.org/en/posa/high-performance-networking-in-chrome.html, DNS-prefetch may not work for proxied sites, especially when it uses [Built-in Asynchronous DNS](https://plus.google.com/+FrancoisBeaufort/posts/JXcvYw1yCkH) rather than system one.

myleshorton: The other possibility that comes to mind @fffw and @oxtoacart is to add sites to the PAC file that we verify are not blocked and have them go **DIRECT**, i.e. the inverse of what we currently do. That would just offload them back to the browser, which is just insanely optimized, keeping things as light as possible on the Lantern side.

myleshorton: I also haven't had a chance to look at the ocde at all, so apologies if that's what you already do!

fffw: Yeah I've thought about it, but wouldn't there be too many items in it? If a site is blocked afterwards, Lantern will not notice it.

myleshorton: For a given session it will only test once right? I.e., if it decides a site is unblocked, it doesn't keep re-checking right? If so then I don't think putting it in the PAC file would be any different as long as it doesn't persist across sessions.

fffw: Yeah you are right, my brain stuck. A PAC file can be changed dynamically. Will do it right away.

fffw: Oh unfortunately through my testing browser (at least Chrome and Firefox) will not retrieve PAC file again until next launch, or when network status changed. This makes dynamically changed PAC nearly useless for those who keeps browser window open (including me).

fffw:
Add the logic of update PAC file if no errors encountered while accessing a site directly.
Chrome and Firefox seems happy for this change.

myleshorton:
Cool thanks @fffw. It's looking like we might do a mini beta 3 today and move this to beta 4 once the code review is all set. How are you feeling on this one @oxtoacart?

fffw:
Ohhh I just found that we can't assume any site to be unblocked and ask browser to bypass Lantern.

User may have connected VPN when Lantern is running. After the VPN disconnected, those blocked sites added to PAC file previously will fail to open, which makes user(yeah it's me) confusing.

Maybe we can leave the choice of  auto detecting or not to user. When Lantern is running, it can be easily switched among 3 states: not proxy at all / proxy based on list (including customised ones)  / auto detection (with small performance penalty).

@myleshorton @oxtoacart How do you think?

oxtoacart:
Okay, this LGTM.  I'm going to go ahead and merge.

myleshorton:
Great point @fffw! Any way we can detect a VPN?

oxtoacart:
I still think we just need to bite the bullet and do an extension.  Then this problem goes away.

atavism:
> Maybe we can leave the choice of auto detecting or not to user. When Lantern is running, it can be easily switched among 3 states: not proxy at all / proxy based on list (including customised ones) / auto detection (with small performance penalty).

Yeah, the settings screen is going to be re-enabled with https://github.com/getlantern/lantern/issues/2268 which is just about finished. As mentioned yesterday, that's an ideal place to include an option to turn auto-detection on/off.

fffw:
Thanks for reviewing @oxtoacart 

I think the chance having Lantern open and VPN connected in same time is small. And user can always relaunch Lantern to resolve it (hope user will do it by instinct).

myleshorton:
Thanks @fffw -- it's been working great in my testing!

PR #116, opened 2015-03-05
Better logging for geolookup errors

oxtoacart:
This doesn't fix getlantern/lantern#2285, but it should help diagnose them.

### DIFF
--- a/src/github.com/getlantern/flashlight/geolookup/geolookup.go
+++ b/src/github.com/getlantern/flashlight/geolookup/geolookup.go
@@ -1,6 +1,7 @@
 package geolookup
 
 import (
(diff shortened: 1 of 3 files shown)


xiam:
Hey @oxtoacart, I noticed that we're targeting the valencia branch, should we continue using it as our development target?

oxtoacart:
@xiam Yes.  I've made valencia the default branch now.  We'll keep devel for supporting the old flashlight used by the Java Lantern version.

xiam:
Got it @oxtoacart, looks good to me!

oxtoacart:
@xiam Thanks for reviewing!

PR #119, opened 2015-03-10
extract prefix and truncate messages to 100 chars

fffw:
This resolves getlantern/lantern#2287

### DIFF
--- a/src/github.com/getlantern/flashlight/logging/logging.go
+++ b/src/github.com/getlantern/flashlight/logging/logging.go
@@ -7,10 +7,9 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
+	"strings"
 	"sync"
 	"time"
-
-	"github.com/getlantern/rotator"
 
 	"github.com/getlantern/appdir"
 	"github.com/getlantern/flashlight/config"
@@ -19,6 +18,7 @@ import (
 	"github.com/getlantern/go-loggly"
 	"github.com/getlantern/golog"
 	"github.com/getlantern/jibber_jabber"
+	"github.com/getlantern/rotator"
 	"github.com/getlantern/waitforserver"
 	"github.com/getlantern/wfilter"
 )
@@ -174,10 +174,22 @@ func (w logglyErrorWriter) Write(b []byte) (int, error) {
 		"timeZone":  w.tz,
 		"version":   w.versionToLoggly,
 	}
-
+	fullMessage := string(b)
+	message := fullMessage
+	// Loggly can't index fields with more than 100 characters
+	if len(fullMessage) > 100 {
+		message = fullMessage[0:100]
+	}
+	pos := strings.IndexRune(message, ':')
+	if pos == -1 {
+		pos = 0
+	}
+	prefix := message[0:pos]
 	m := loggly.Message{
-		"extra":   extra,
-		"message": string(b),
+		"extra":        extra,
+		"locationInfo": prefix,
+		"message":      message,
+		"fullMessage":  fullMessage,
 	}
 
 	err := w.client.Send(m)
--- a/src/github.com/getlantern/flashlight/logging/logging.go
+++ b/src/github.com/getlantern/flashlight/logging/logging.go
@@ -28,7 +28,7 @@ const (
 )
 
 var (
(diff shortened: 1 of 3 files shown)


src/github.com/getlantern/flashlight/logging/logging.go
```diff
@@ -174,10 +174,22 @@ func (w logglyErrorWriter) Write(b []byte) (int, error) {
 		"timeZone":  w.tz,
 		"version":   w.versionToLoggly,
 	}
-
+	fullMessage := string(b)
+	message := fullMessage
+	// Loggly can't index fields with more than 100 characters
```
oxtoacart: Small nitpick, but I think Loggly can still index these fields, it just doesn't generated facets for them.

oxtoacart:
Looks good.  The only thing I would add is a little unit test for the truncation logic to make sure it behaves as expected and also to illustrate what it's supposed to do.

fffw:
Yeah a unit test is definitely required. I'll add it

src/github.com/getlantern/flashlight/logging/logging_test.go
```diff
@@ -0,0 +1,36 @@
...
+		assert.Equal(t, "test: short message\n", result["message"])
+	}
+
+	buf.Reset()
+	longMsg := "really l" + strings.Repeat("o", 100) + "ng message"
+	log.Error(longMsg)
+	if assert.NoError(t, json.Unmarshal(buf.Bytes(), &result), "Unmarshal error") {
+		assert.Equal(t, "test", result["locationInfo"])
+		assert.Equal(t, ("test: " + longMsg)[0:100], result["message"], "message should be truncated to 100 characters")
+	}
```
oxtoacart: I like how you're testing through the loggly API!  I thought of a couple more things that are probably good to do:

1. Add this test to `testpackages.txt`

2. I think we can use better logic for really long messages.  

Take for example this error: `Mar 09 22:00:27.900 - enproxy: Dest: ui.getlantern.org:443    ProxyHost: fl-nl-20150226-007.getiantem.org    Unable to issue read request: Error reading response from proxy: unexpected EOF`.

Your logic takes the first 100 bytes, which is `enproxy: Dest: ui.getlantern.org:443    ProxyHost: fl-nl-20150226-007.getiantem.org    Unable to is`

It would also be useful to take chunks from the tail, terminating at the colons.  For the example, this would be:

```
unexpected EOF
Error reading response from proxy: unexpected EOF
Unable to issue read request: Error reading response from proxy: unexpected EOF
```

These would be very nice for grouping since they're complete messages and will often match similar messages.

fffw: @oxtoacart Great idea!
After consulting current logs, I think keeping last two chunks is good enough for us to group logs of same reason.

fffw: @oxtoacart Great idea!
After consulting current logs, I think keeping last two chunks is good enough for us to group logs of same reason.

oxtoacart: Last 2 is a good starting point.  I just figured that keeping different sets of chunks would be useful in case some of them include variable data (e.g. hostname).

src/github.com/getlantern/flashlight/logging/logging.go
```diff
@@ -174,10 +174,34 @@ func (w logglyErrorWriter) Write(b []byte) (int, error) {
...
+	pl := len(parts)
+	switch pl {
+	case 1:
+		message = ""
+	case 2:
+		message = parts[1]
+	default:
+		message = parts[pl-2] + ":" + parts[pl-1]
+	}
+	message = strings.Trim(message, " \n")
```
oxtoacart: You could use strings.TrimSpace here to trim all leading and trailing whitespace.

oxtoacart:
Other than maybe using strings.TrimSpace (and checking for that in your unit test) this LGTM.

fffw:
Thanks for pointing out @oxtoacart. Unit test already covered TrimSpace as fullMessage always has trailing line feed, and leading whitespace was removed at test file L47.

oxtoacart:
Nice work!

PR #128, opened 2015-03-14
Multifronting valencia

aranhoide:
[**Note:** The diff here is quite unworkable because of the AWS and DNSimple bindings.  You may want to review (and comment on) https://github.com/getlantern/flashlight-build/pull/129 instead, which is the same as this without these libraries.  Please don't merge that one, though.]

This includes the changes in the multifronting PR, plus changes to peerscanner so it can register servers in Cloudfront and in DNSimple (Cloudfront requires that its "distributions" point to domain names, not IPs).

In addition to Cloudflare entries, DNSimple individual (i.e. not round robin) entries are created for all servers.  In addition, Cloudfront distributions are created for all servers that report supporting Cloudfront.  When we receive a register request from a server that has a Cloudfront distribution set up (these take up to 15 minutes to get ready) we add these entries to the DNSimple roundrobin.  We have a bunch of manually created distributions in Cloudfront that point to the DNSimple round robins.

I have tested these with the DO droplets `ps-ara` (ip 128.199.40.248, registered as `ps-test.getiantem.org`), `fl-nl-ara1` (188.166.38.243, serving via HTTPS [edit: as of this writing, this will not really work with Cloudfront, because it's pickier than Cloudflare about self-signed certs]) and `fl-nl-ara2` (178.62.237.77, serving with `-unencrypted` on port 80).  I tested that these get registered and unregistered (only from the 'groups', particular entries are not unregistered).  I had tested the proxying for the multifront PR; nothing has changed in that aspect.

There will be a twin PR for the lantern_aws changes required to launch/update these.


myleshorton:
This is kind of impossible to sort through with the AWS-SDK included in the diff. Any way to just add that in a separate PR?

aranhoide:
@myleshorton : I've just added https://github.com/getlantern/flashlight-build/pull/129 , which is this PR with the AWS and DNSimple bindings removed.  Feel free to comment in that one (but when the time comes to merge, please merge this one).

myleshorton:
OK, I'm merging this, but I do think we should go back in and tidy up soon.

aranhoide:
Thanks!

I agree this needs tidying up, and unit tests.  If I don't get to it before, this Friday is Quality Day.

PR #127, opened 2015-03-14
getlantern/lantern#2306 Added full list of fallback proxies as chained s...

oxtoacart:
...ervers

PR #129, opened 2015-03-15
Multifront valencia nodeps

aranhoide:
**DON'T MERGE THIS**: this is just the same as https://github.com/getlantern/flashlight-build/pull/128 except the AWS and DNSimple bindings have been removed to make the diff more workable.  Feel free to comment here, but please merge https://github.com/getlantern/flashlight-build/pull/128, not this, when you're done.

Removing these libraries broke travis; don't worry about that, the original PR passes CI tests.


myleshorton:
OK thanks @aranhoide! Sorry not sure what the best process is for something like that, but seems to have come up a few times, so let's just revisit in the aftermath. I'm actually thinking it might make sense to remove CloudFlare entirely and to possibly re-introduce it if we don't get blocked, but that should be an easy change.

src/github.com/getlantern/enproxy/proxy.go
```diff
@@ -169,11 +175,15 @@ func (p *Proxy) handleWrite(resp http.ResponseWriter, req *http.Request, lc *laz
...
-		// through (e.g.) DNS round robin.
-		resp.Header().Set(X_ENPROXY_PROXY_HOST, p.Host)
+	host := ""
+	if p.HostFn != nil {
+		host = p.HostFn(req)
+	}
+	if host == "" {
+		host = p.Host
+	}
+	if host != "" {
```
myleshorton, line 185: Maybe keep the comment from the previous line 173 here? I feel like some more comments here would be handy in general, like why do we preference the HostFn if it exists over the Host?

src/github.com/getlantern/flashlight/server/server.go
```diff
@@ -125,15 +149,25 @@ func (server *Server) ListenAndServe() error {
...
+func (server *Server) register(updateConfig func(func(*ServerConfig) error)) {
+	supportedFronts := make([]string, 0, len(frontingProviders))
+	for name := range frontingProviders {
+		supportedFronts = append(supportedFronts, name)
+	}
 	for {
 		server.cfgMutex.RLock()
 		baseUrl := server.cfg.RegisterAt
+		var port string
+		if server.cfg.Unencrypted {
```
aranhoide, line 166: I just realised we should really get the port from the `-addr` value, but then we'd need to make `-unencrypted` take another port (and perhaps another host).  I'll think about this in the context of implementing dual listening.  In the meanwhile, this will work for our own fallbacks, which at the moment are the only ones peerscanner is registering.

src/github.com/getlantern/enproxy/proxy.go
aranhoide, line 185: The purpose of HostFn (and Host) is documented where the fields are defined, but I'll add this comment back since it can't hurt, I guess.

The Host is only kept around for backwards compatibility, since enproxy is a library we may expect other people to use.  I can't see why anyone would use both Host _and_ HostFn.  HostFn is given preference because if you use that the only reason I can think about for providing a Host too is if you've forgotten to remove it.  I'll add some comments to this effect too.

aranhoide, line 185: @myleshorton: I've just added some comments to the actual PR:

https://github.com/getlantern/flashlight-build/commit/6feb9b51b62abc3a40f14121c871d1eeb72bd893

src/github.com/getlantern/flashlight/flashlight.go
```diff
@@ -261,7 +260,14 @@ func runServerProxy(cfg *config.Config) {
...
 
-	err = srv.ListenAndServe()
+	err = srv.ListenAndServe(func(update func(*server.ServerConfig) error) {
+		err := config.Update(func(cfg *config.Config) error {
+			return update(cfg.Server)
+		})
+		if err != nil {
+			log.Errorf("Error while trying to update: %v", err)
+		}
+	})
```
myleshorton, line 270: It seems to me like it might be better to just make the srv.register method public and call that after ListenAndServe, as the updateConfig func is only used in register, and we only really want to register after we're 100% sure the server is up (there's a minor, unlikely to be hit race condition with the go routine in ListenAndServe with registering before we're actually serving as far as I can tell).

Does ListenAndServe block though?

src/github.com/getlantern/flashlight/server/server.go
```diff
@@ -27,17 +27,42 @@ const (
...
+		if entry == value {
+			return true
+		}
+	}
+	return false
+}
+
+func hasHeader(req *http.Request, name string) bool {
+	return req.Header[name] != nil
+}
```
myleshorton, line 57: Seems like the above two methods would ideally be in some httputil file for general use.

src/github.com/getlantern/flashlight/flashlight.go
oxtoacart, line 270: ListenAndServe does block.

ListenAndServe only registers after successfully calling Listen(), which is really the thing that could fail.  Serve simply dispatches incoming requests.  There's no race condition because as soon as Listen was successful, the server will accept connections.  There might just be a slight delay before it reads from the connection.

src/github.com/getlantern/flashlight/server/server.go
```diff
@@ -27,17 +27,42 @@ const (
...
+		},
+		"cloudfront": func(req *http.Request) bool {
+			return hasHeader(req, "X-Amz-Cf-Id") || headerMatches(req, "User-Agent", "Amazon Cloudfront")
+		},
+	}
 )
 
+func headerMatches(req *http.Request, name string, value string) bool {
+	for _, entry := range req.Header[name] {
+		if entry == value {
```
myleshorton, line 48: Might also make sense to just always test lowercased versions since HTTP headers are case-insensitive.

src/github.com/getlantern/enproxy/proxy.go
myleshorton, line 185: Cool thanks! I _think_ at this point no one else is using enproxy, so we can probably get away with breaking some backwards compatibility, but this looks fine.

src/github.com/getlantern/flashlight/server/server.go
aranhoide, line 48: Do you mean of header names or values?  Golang's http library normalizes the header names, as described in the comment above.

src/github.com/getlantern/flashlight/server/server.go
aranhoide, line 57: Do we have something like that, or other methods to put there?  Creating a httputil file/lib only for these two seems overkill.

src/github.com/getlantern/peerscanner/main.go
```diff
@@ -69,122 +107,238 @@ func main() {
...
+		g := cflGroups[name]
 		if g == nil {
 			g = make(map[string]*cloudflare.Record, 1)
-			groups[name] = g
+			cflGroups[name] = g
 		}
 		g[r.Value] = &r
 	}
 
+	// Collect round-robin entries in Cloudflare
```
myleshorton, line 185: Nit: in DNSSimple

src/github.com/getlantern/flashlight/server/server.go
myleshorton, line 48: I was thinking the values, but come to think of it I think only HTTP header names are case insensitive, so all good.

src/github.com/getlantern/flashlight/server/server.go
myleshorton, line 57: We don't have that no. Fine to leave for now, but we should probably start building something like that up sometime soon.

src/github.com/getlantern/peerscanner/main.go
aranhoide, line 185: Thanks!  Fixed in the main PR.

src/github.com/getlantern/flashlight/flashlight.go
myleshorton, line 270: OK cool.

src/github.com/getlantern/peerscanner/web.go
```diff
@@ -51,36 +56,46 @@ func startHttp() {
 // register is the entry point for peers registering themselves with the service.
 // If peers are successfully vetted, they'll be added to the DNS round robin.
 func register(resp http.ResponseWriter, req *http.Request) {
-	name, ip, port, err := getHostInfo(req)
-	if err == nil && port != 443 {
-		err = fmt.Errorf("Port %d not supported, only port 443 is supported", port)
+	name, ip, port, supportedFronts, err := getHostInfo(req)
+	if err == nil && !(port == "80" || port == "443") {
+		err = fmt.Errorf("Port %d not supported, only ports 80 and 443 are supported", port)
```
myleshorton, line 61: Wait so what's the rationale for allowing registration over port 80?

src/github.com/getlantern/peerscanner/web.go
```diff
@@ -51,36 +56,46 @@ func startHttp() {
...
+	connectionRefused := false
+	timedOut := false
 
+	h := getOrCreateHost(name, ip, port)
+	online, connectionRefused, timedOut = h.status()
 	if online {
 		resp.WriteHeader(200)
 		fmt.Fprintln(resp, "Connectivity to proxy confirmed")
+		if (supportedFronts & cloudfrontBit) == cloudfrontBit {
+			h.initCloudfront()
```
myleshorton, line 84: This is the bit that takes ~15 minutes, is that right?

src/github.com/getlantern/peerscanner/web.go
aranhoide, line 61: Peerscanner tests the servers by connecting to them directly by IP.  If we allow the hop between the CDN and the flashlight server to be unencrypted (and currently this is the only way I could get flashlight to work with Cloudfront so far), then the flashlight server will be listening in port 80.  The hop between the client and the CDN is of course always encrypted, over port 443, but peerscanner skips that.

aranhoide, line 61: I've sent an email to bns-ops on this issue.  Eventually we'll need to either get a CA-signed cert so Cloudfront will connect to us via HTTPS on 443, or have flashlight servers listen dually (port 443 will always be needed for clients that use the proxy without domain fronting.)

src/github.com/getlantern/peerscanner/web.go
aranhoide, line 84: It takes ~15 minutes since you create a Cloudfront distribution (CDN domain) until it is ready.   But this call never blocks for 15 minutes, if that's what you're asking.

This call just makes sure Cloudfront is/gets initialized, moving things forward as necessary.  It doesn't always create a Cloudfront distribution, and even when it does, it returns rather quickly.  The DNSimple round robin records will be added the next time this is called after the distribution is ready.

src/github.com/satori/go.uuid/LICENSE
```diff
@@ -0,0 +1,20 @@
+Copyright (C) 2013-2015 by Maxim Bublis <b@codemonkey.ru>
```
aranhoide: Whoops, I'll remove this too from this branch.  It is only used by tests anyway.

src/github.com/getlantern/peerscanner/web.go
myleshorton, line 61: Makes sense. Doesn't that mean we need to also listen on port 80 now? Sorry I've been getting hammered with email so missed the discussion.

src/github.com/getlantern/peerscanner/web.go
```diff
@@ -183,3 +218,15 @@ var fallbackIPPrefixes = []string{
...
+	ls := req.Form[name]
+	if len(ls) == 0 {
+		return ""
+	}
+	if len(ls) > 1 {
+		// But we still allow it for robustness.
+		log.Errorf("More than one '%v' provided in form: %v", name, ls)
+	}
+	return ls[0]
+}
```
myleshorton, line 232: Another candidate for an httputils.go file, but not biggie

oxtoacart, line 232: [gorilla](http://www.gorillatoolkit.org/) or something similar may already provide a lot of this kind of functionality.

src/github.com/getlantern/peerscanner/web.go
oxtoacart, line 61: I don't see HTTPS to our servers as a huge priority since we're allowing untrusted peers to register too.  Plus, the CDN is able to MITM stuff anyway, so as far as MITM'ing goes, there's lots of holes.  Best for users to make sure to use HTTPS end-to-end when possible.

src/github.com/getlantern/peerscanner/main.go
```diff
@@ -69,122 +107,238 @@ func main() {
...
-	groups := make(map[string]map[string]*cloudflare.Record, 0)
+	dists, err := cfr.ListDistributions(cfrutil)
+	if err != nil {
+		return nil, fmt.Errorf("Unable to load cloudfront distributions: %v", err)
+	}
+	log.Debugf("Loaded %d existing distributions", len(dists))
 
-	addToGroup := func(name string, r cloudflare.Record) {
+	// Collect round-robin entries in Cloudflare
+	cflGroups := make(map[string]map[string]*cloudflare.Record, 0)
```
myleshorton, line 174: It seems like there's an opportunity for refactoring here. Like what about a CloudfrontRecords and CloudFlareRecords struct, each of which implements an interface with add, remove, etc, and that just looks up all the records when it's created, something like that? Seems like that could simplify things a bit, but this certainly looks like it works for the purposes of getting this out the door.

src/github.com/getlantern/flashlight/README.md
```diff
@@ -45,7 +45,7 @@ Usage of flashlight:
   -portmap=0: try to map this port on the firewall to the port on which flashlight is listening, using UPnP or NAT-PMP. If mapping this port fails, flashlight will exit with status code 50
   -registerat="": base URL for peer DNS registry at which to register (e.g. https://peerscanner.getiantem.org)
   -role="": either 'client' or 'server' (required)
-  -server="": FQDN of flashlight server when running in server mode (required)
+  -frontfqdns="": YAML string representing a map from the name of each front provider to a FQDN that will reach this particular server via that provider (e.g. '{cloudflare: fl-001.getiantem.org, cloudfront: blablabla.cloudfront.net}')
```
atavism, line 48: Just curious: Is passing a YAML string somehow preferable to passing a string with '=' as the character delimiter? i.e. 
`-frontfqdns cloudflare=fl-001.getiantem.org cloudfront=blablabla.cloudfront.net`? That's less to type and just as easy to parse.

src/github.com/getlantern/peerscanner/cfr/cfr_test.go
```diff
@@ -0,0 +1,61 @@
...
+	assert.True(t, strings.HasSuffix(dist.Domain, ".cloudfront.net"), "Domain should be a .cloudfront.net subdomain, not '"+dist.Domain+"'")
+	err = RefreshStatus(cfr, dist)
+	assert.NoError(t, err, "Should be able to refresh status")
+	// Just check that Status stays a valid one.  Checking that it eventually
+	// gets refreshed to "Deployed" would take a few minutes, and thus is out
+	// of the scope of this unit test.
+	assert.Equal(t, "InProgress", dist.Status, "New distribution should have Status: \"InProgress\" even after refreshing right away")
+	assert.NoError(t, counter.AssertDelta(0), "All file descriptors should have been closed")
+}
+
```
atavism, line 54: I ran `go test` without having CFR_KEY or CFR_ID set. It might be nice to display a warning instead of proceeding ahead and trying to deference a nil pointer? At least a more suitable error message would be helpful in this scenario.

```
--- FAIL: TestCreateAndRefresh (0.13s)
        Location:       cfr_test.go:42
    Error:      No error is expected but got The security token included in the request is invalid.
    Messages:   Should be able to create distribution

panic: runtime error: invalid memory address or nil pointer dereference [recovered]
    panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xb code=0x1 addr=0x0 pc=0x6719c]

goroutine 12 [running]:
testing.func·006()
    /usr/local/go/src/testing/testing.go:441 +0x181
github.com/getlantern/peerscanner/cfr.TestCreateAndRefresh(0xc208296ea0)
    /Users/todd/Dev/go/src/github.com/getlantern/flashlight-build/src/github.com/getlantern/peerscanner/cfr/cfr_test.go:43 +0x11ac
testing.tRunner(0xc208296ea0, 0x57b458)
    /usr/local/go/src/testing/testing.go:447 +0xbf
created by testing.RunTests
    /usr/local/go/src/testing/testing.go:555 +0xa8b
```

src/github.com/getlantern/peerscanner/web.go
aranhoide, line 61: In order to sit behind Cloudfront, flashlight servers currently need to listen for unencrypted connections (e.g. in port 80).

In order to serve to get mode clients directly via IP (i.e., no domain fronting), flashlight servers would need to listen through HTTPS.

Currently, flashlight servers can only listen through either one or the other.  So if we got blocked right now, we'd need different pools of servers for both purposes.  This is pretty much what we have now with flashlight servers vs old-style fallbacks.  It would be more efficient to serve in both ways from the same pool of servers.

The question is: in the long run, do we prefer to keep the hop between the CDN and flashlight servers encrypted?  If so, I think we need to get a wildcard CA-signed cert for (e.g.) flashlightproxy.org, so Cloudfront will serve HTTPS to them.

If we are happy with having that hop unencrypted (saves some CPU; may have security repercussions, but out of our threat model), then we should change flashlight servers so they can listen on both ports.

I think both changes are rather easy to implement, so it's just a matter of deciding whether we want the CDN<->flashlight server hop encrypted or not.

myleshorton, line 61: Yeah the other big factor here is just that we already tunnel HTTPS end-to-end. So the only traffic that's effectively in the clear from CDN -> Lantern (peers or no) is HTTP traffic. Since that's obviously in the clear anyway, I don't think we care at all about the encryption from CDN -> Lantern. In fact, since we're not going to be shipping certs to peer proxies any time soon, there's also no real way we could do that as far as I can tell.

src/github.com/getlantern/peerscanner/web.go
myleshorton, line 84: Right -- I was just asking if that's where the 15 minute Cloudfront distribution kickoff takes place, not that it would ever block for 15 minutes (good god!).

src/github.com/getlantern/peerscanner/web.go
oxtoacart, line 61: > In order to serve to get mode clients directly via IP (i.e., no domain fronting), flashlight servers would need to listen through HTTPS.

That's not the only thing that would need to change, since domain-fronting uses the enproxy protocol, which is unnecessary for chained proxying.  Flashlight (now Lantern) can be modified to support the use of chained proxying using the chained package (which already has a server component).  It would also need to implement the Apache HTTP server mimicry that the fallbacks currently do, along with the auth_token checking.

aranhoide, line 61: Are you already working on this, or is this something I could try my hand at while you're out?  (Sorry about throwing work mail at you BTW.)

src/github.com/getlantern/peerscanner/main.go
myleshorton, line 174: What do you think about this @aranhoide?

src/github.com/getlantern/peerscanner/host.go
```diff
@@ -80,33 +87,75 @@ func (h *host) String() string {
...
+			h.dspGroups[country] = &dspGroup{subdomain: country}
+		}
+	} else {
+		log.Errorf("Somehow adding peer host? %v (%v)", name, ip)
+	}
+
+	return h
+}
+
+func (h *host) resetProxiedClient(port string) {
```
myleshorton, line 129: A godoc comment here would be super handy I think.

src/github.com/getlantern/peerscanner/host.go
```diff
@@ -80,33 +87,75 @@ func (h *host) String() string {
...
+
+	var dial func(addr string) (net.Conn, error)
+	if port == "80" {
+		dialer := net.Dialer{Timeout: dialTimeout}
```
myleshorton, line 133: Could scope this a little tighter and put it down two lines

src/github.com/getlantern/peerscanner/main.go
aranhoide, line 174: Yes, definitely some polymorphism would help, both in groups and in DNS records.  I kept it straightforward, at the cost of some repetition, for the benefit of getting it out the door, as you say.  Also, some things stick out, like Cloudfront's relationship with DNSimple.

Another opportunity for refactoring would be to extract pure functions that depend only on their arguments and return a list of actions to be done.  These functions would be easily unit testable, and the parts that (i) actually get the data from cfl/cfr/dsp to feed to these functions, and (ii) perform the work returned by them, would be dumb and straightforward enough that the unit tests for the cfl/cfr/dsp libraries would give us enough confidence that these parts work.  This approach would be easiest to apply in the case of `main.loadHosts`.  Doing the same for the main loop of hosts would be trickier without obfuscating the code and negating some of the niceties of inversion of control allowed by golang channels, but I guess it could be done.  Maybe a project for Quality Day, since this would increase unit testing and make the other refactorings you mention less onerous to QA.

src/github.com/getlantern/peerscanner/host.go
aranhoide, line 133: Then we would be creating a dialer each time `dial` is called.  This way, it gets reused for the lifetime of the `proxiedClient`, saving some allocation and garbage collection.  Maybe this savings will never matter, but it's easier and cheaper to "optimize" this than to determine whether it really matters. :)  

`dialer` is scoped to the `if` statement anyway, where the only thing that happens is defining `dial`.

src/github.com/getlantern/flashlight/README.md
aranhoide, line 48: (tl;dr Not a big deal either way, I guess.)

I don't know whether golang will parse those two "words" as the value of `-frontfqdns` without surrounding them with quote marks.  Anyway, we were already including YAML in flashlight (this frontfqdns dict itself will be saved as YAML in the config), it takes one line to parse, it looks less like something that could be confused with `-argument=val` bash syntax, and it's a well-known format so we can describe the format with two words in the help string.

src/github.com/getlantern/peerscanner/host.go
aranhoide, line 129: Added:

https://github.com/getlantern/flashlight-build/commit/53ceb195d14c100321eb38e68b5c94dadf8c9827

Note that this thing about being flexible with what ports servers could support made sense mostly in the context of supporting already existing peers.  Now that we control all the flashlight servers that are allowed to register with peerscanner, we could just update them all to use the same port(s) and assume that in peerscanner, thus removing a bit of unneeded complexity.

src/github.com/getlantern/peerscanner/cfr/cfr_test.go
aranhoide, line 54: Good point.  Fixed:

https://github.com/getlantern/flashlight-build/commit/6717dd890fe6702c2b00af8422ece1c919458da5

This was only an issue with tests.  Peerscanner itself was already making these checks and giving more helpful error messages.

src/github.com/getlantern/peerscanner/web.go
oxtoacart, line 61: I haven't done anything with this, no.  Is this critical path right now?

src/github.com/getlantern/peerscanner/host.go
myleshorton, line 133: That's a legitimate point. I think in practice it's really unlikely that allocation would show up on the radar, but I'm ok with it.

src/github.com/getlantern/peerscanner/web.go
myleshorton, line 61: I think the direct IP proxies are critical path at this point yeah.

aranhoide, line 61: But if you haven't started I'll give it a shot.
O 17 Mar, 2015 06:45, "myleshorton" notifications@github.com escribiu:

oxtoacart, line 61: We already have direct ip-based proxies - https://github.com/getlantern/flashlight-build/pull/127

My question was, is migrating the back-end for these from the old Java-based program to the new Go-based program critical path?

aranhoide, line 61: I thought it was, but I don't think so anymore.  More details on bns-ops.

myleshorton, line 61: This is a related issue, but I think we ultimately need to go back to testing registered peers through the full domain-fronted stack instead of by direct IP. Peerscanner should be in charge of or at least should directly influence cloud.yaml. The way it stands now if a specific domain front stopped working, peerscanner would never know, and the cloud.yaml would not get updated. If we did the full domain fronted proxy call instead, peerscanner could automatically update the cloud.yaml to give clients a shot at continuing to function. 

On a related note, we should also start including the cloud.yaml URL in the cloud.yaml itself, as ideally we'd store it on an unblocked mirror and could change that mirror easily. That actually calls for going back to direct calls to the cloud.yaml and using it directly if it's not blocked to bypass the potential point of failure in domain-fronted calls.

Peerscanner will eventually also need to directly affect cloud.yaml by setting the percentage of traffic going to peers based on how many peers there are versus fallbacks at any given time.

I don't think there's anything we need to change immediately, but I'll ticket having peerscanner go back to how it used to do domain-fronted checks for fallbacks so it can update cloud.yaml directly.

src/github.com/getlantern/peerscanner/host.go
```diff
@@ -57,15 +59,20 @@ type status struct {
 type host struct {
 	name        string
 	ip          string
-	record      *cloudflare.Record
+	port        string
+	cflRecord   *cloudflare.Record
```
myleshorton, line 63: Can you update the comment for this struct while we're at it to include CloudFront?

src/github.com/getlantern/peerscanner/host.go
myleshorton, line 133: Looking at this again I think that's not a good premature optimization (some would call that an oxymoron, but I'm actually with you that cheap, no brainer optimizations in potentially performance critical code are fine). Creating a single dialer that itself stores very little memory isn't going to do much -- to me the clarity and correctness of the code should win here. But then again, this is a nit to begin with!

src/github.com/getlantern/peerscanner/host.go
```diff
@@ -407,11 +484,32 @@ func (h *host) isAbleToProxy() (bool, bool, error) {
 }
 
 func (h *host) doIsAbleToProxy() (bool, bool, error) {
+	if h.port == "" {
```
myleshorton, line 487: This could use a comment. This is to support old clients that don't report the port, is that correct? So in that case we try both 80 and 443?

src/github.com/getlantern/peerscanner/host.go
myleshorton, line 133: I would actually even take this further -- i.e. why are we even storing state with proxiedClient at all? Why not just create it when we need it with a newClient call, passing the port so it can figure out the correct dialer and whatnot?

aranhoide, line 133: I put dialer in the tighter scope, because it's a tradeoff between infinitesimals either way.

https://github.com/getlantern/flashlight-build/commit/7f05a1a81ed3ea753ee093435ea572299e42fb6e

TBH I don't see any correctness hazard or clarity problem here.  On that respect, `dialer` might as well be global.  It'd be no worse than `(cfl|cfr|dsp)util`, and those are used less often.

We are not storing any state (as in, anything that changes) in proxiedClient.  resetProxiedClient does what (IIUC) you want newClient to do, except, as a slight convenience, it also sets the client to the host when it could be created successfully.  So we can say:

```
h.resetProxiedClient(port)
```

instead of

```
pc := newProxiedClient(h, port)
if pc != nil {
    h.proxiedClient = pc
}
```

But that's not a big deal either and I can change it too if you want.

myleshorton, line 133: > TBH I don't see any correctness hazard or clarity problem here. On that respect, dialer might as well be global. It'd be no worse than (cfl|cfr|dsp)util, and those are used less often.

I just think state should be avoided wherever possible, and it doesn't seem like it buys us anything here -- I'm certainly not condoning the other globals by questioning this one!

> We are not storing any state (as in, anything that changes) in proxiedClient. 

Well, proxiedClient itself is updated with it's new state -- each proxied client itself is immutable I suppose sure.

> resetProxiedClient does what (IIUC) you want newClient to do, except, as a slight convenience, it also sets the client to the host when it could be created successfully. 

I would actually do the following:
1. Remove proxiedClient from the host struct
2. Kill the doIsAbleToProxy method entirely
3. Create a new method for creating the client that just gets called when we need the client, which is currently further down in reallyDoIsAbleToProxy

That also allows us to only deal with the client if and when we actually need it -- i.e. once the raw connection test has happened. It's not a huge deal, but otherwise we're doing all this to me confusing gymnastics just to reset the proxiedClient to have the correct dialer, which is hardly an optimization to begin with.

``` go
func (h *host) newProxiedClient(port string) http.Client {
    var dial func(addr string) (net.Conn, error)
    if port == "80" {
        dialer := net.Dialer{Timeout: dialTimeout}
        dial = func(addr string) (net.Conn, error) {
            return dialer.Dial("tcp", h.ip+":80")
        }
    } else if port == "443" {
        dial = func(addr string) (net.Conn, error) {
            return tlsdialer.DialWithDialer(&net.Dialer{
                Timeout: dialTimeout,
            }, "tcp", h.ip+":443", true, &tls.Config{
                InsecureSkipVerify: true,
                // Cache TLS sessions
                ClientSessionCache: tls.NewLRUClientSessionCache(1000),
            })
        }
    } else {
        log.Errorf("Unsupported port: %v", port)
        return
    }

    return &http.Client{
        Transport: &http.Transport{
            Dial: func(network, addr string) (net.Conn, error) {
                return enproxy.Dial(addr, &enproxy.Config{
                    DialProxy: dial,
                    NewRequest: func(upstreamHost string, method string, body io.Reader) (req *http.Request, err error) {
                        return http.NewRequest(method, "http://"+h.ip+"/", body)
                    },
                    OnFirstResponse: func(resp *http.Response) {
                        h.reportedHostMutex.Lock()
                        h.reportedHost = resp.Header.Get(enproxy.X_ENPROXY_PROXY_HOST)
                        h.reportedHostMutex.Unlock()
                    },
                })
            },
            DisableKeepAlives: true,
        },
        Timeout: requestTimeout,
    }
}
```

myleshorton, line 133: This is outside the scope of this PR probably, but on the topic of state I also don't understand why the pause and reset scheme is desirable. Why not just completely remove a host when it's apparently died and then add it back in when we see it again? We just have to worry about getting more right the way it is now as far as I can tell.

aranhoide, line 133: I assumed that the point of tls.NewLRUClientSessionCache(1000) was that it was worth reusing across calls to `dial` and that's why it was created at a scope such that it would remain the same across such calls.  Isn't that the thing whereby you avoid having to redo the whole TLS handshake on every connection?

Also, I made this thing uglier trying to accomodate the ports that servers report.  We can probably mandate port 80 (or 443 if testing with domain fronting), and dispense with that (and with servers reporting port altogether).  Then the proxiedClient can be created on host creation and stay unchanged for its lifetime, which is less confusing and (if using TLS) keeps the cache working.

Re: pause, in principle I agree, but I haven't given it much thought in the context of this PR.  I don't suppose that's urgent to address, so let's hear Ox's rationale when he's back.

src/github.com/getlantern/peerscanner/web.go
aranhoide: I think checking domain fronting providers and the ratio of peers vs fallbacks could be done in a different service, the responsibility of which would be to update cloud.yaml (as opposed to the DNS/CDN registries, which this service would treat as read-only).

Either way, whoever checks fronting providers only needs test the round robin itself, to make sure that it can proxy through _some_ server via each fronting provider.  That would be way cheaper than peerscanner using domain fronting all the time, esp. if we switch to serving domain fronted requests through port 80.

src/github.com/getlantern/peerscanner/host.go
myleshorton, line 133: Oh yeah great point -- TLS session caching. Hmmn...you also could theoretically just store the session cache!

aranhoide, line 133: Yes, I could.  In practice, I think we want to just mandate port 80 for all servers and do away with port reporting and TLS altogether.  We can deploy this PR as is to avoid service interruptions during transition, and remove support for TLS and deploy a cleaned up PR after all servers are updated.

In that case, I think creating the proxiedHost on host creation and never touching it again is just as clear as creating a new one for every check, so I don't care much which we pick.

myleshorton, line 133: Yeah I think that sounds right. I don't see any actual bugs in the PR, so we can probably just merge. I do think we should go back and simplify as much as possible after the merge though, as I think things like this could quickly get out of hand and harder to understand if we don't do cleanups as we go.

Sorry this has taken so long -- peerscanner in particular has changed quite a bit since I last worked on it so took me a bit to get back into it.

PR #130, opened 2015-03-17
avoid breaking out of register loop on PostForm error

aranhoide:


myleshorton:
Nice! Yeah we seem to have quite a lot of servers out of rotation, so would be awesome to deploy this.

aranhoide:
OK, will do.  But first I suppose I should backport this to `devel`, since we were doing releases from there (right, @oxtoacart ?).  Just in case it takes a while for you to read this, I'll just restart the flashlight service on these machines, which should get them back into rotation.

myleshorton:
Oh good point @aranhoide. Are you talking about restarting flashlight on all the machines or just the ones from the syslog? Keep in mind that was just the syslog from _today_ -- the syslogs for prior days are similar. So we're killing something like 20 of these/day.

aranhoide:
Oh wow.

I did restart them all.  A ~dozen of the flashlight servers are not responding to the cloudmaster, I guess because they're overloaded.  I'll look into them.

aranhoide:
Those twelve seem to be hopelessly overloaded.  Connecting to them via ssh times out.  Although those were in Singapore, I'll just kill them and launch new ones in the Netherlands, which IIRC worked better than Singapore for China anyway.

myleshorton:
Awesome thanks!

On Tuesday, March 17, 2015, aranhoide notifications@github.com wrote:

> Those twelve seem to be hopelessly overloaded. Connecting to them via ssh
> times out. Although those were in Singapore, I'll just kill them and launch
> new ones in the Netherlands, which IIRC worked better than Singapore for
> China anyway.
> 
> —
> https://github.com/getlantern/flashlight-build/pull/130#issuecomment-82643634
> .

## 

## 

President
https://www.getlantern.org

PR #132, opened 2015-03-19
Revert "Multifronting valencia"

oxtoacart:
Reverts getlantern/flashlight-build#128

We need to do this because the multifronting-valencia branch contains a bad reverse merge from valencia that wiped out changes in the valencia branch.


myleshorton:
OK, so the plan is to revert, then reverse merge again, and merge back in correct?

oxtoacart:
@myleshorton Plan is perhaps too strong a word for what I was thinking.  The bad merge is actually pretty far back (6 days), so it's not clear to me what is the best approach.  I just figured we should revert the pull that broke at least one thing and perhaps others and rebuild beta 2.

myleshorton:
OK cool yeah it's not clear to me what the best approach is. I do remember there being some merge issue with that particular chunk of code, though, so hopefully it is isolated to that. Based on manual testing and snooping that also seems likely.

PR #133, opened 2015-03-19
getlantern/lantern#2296 Refixed after bad merge

oxtoacart:
Turns out it was just the one line that needed fixing.


myleshorton:
Awesome thanks @oxtoacart -- this seems like the way to go to me!

PR #137, opened 2015-03-20
Pointing at new cloud.yaml that includes chained fallback servers

oxtoacart:
The yaml has already been uploaded.

If this is working, the vast majority of your traffic should look like it has an IP address in NY (not Singapore, not Amsterdam).


myleshorton:
Yup looks good. I do think we want to start deploying direct proxies somewhere closer soon as well, but I'll add that as a separate ticket.

PR #138, opened 2015-03-20
User settings and server-side analytics

atavism:
This includes changes for https://github.com/getlantern/lantern/issues/2268 and https://github.com/getlantern/lantern/issues/769


myleshorton:
Generally looks great @atavism! It looks like this also addresses https://github.com/getlantern/lantern/issues/2268 right? On that front I still think that first window showing the settings should probably show those settings within a Welcome Screen, as otherwise it looks like it came out of nowhere a little bit. So something like (forgive the shoddy colors!):

image

Then the next screen might just have the text and Twitter/FB links but no Welcome to Lantern at the top.

Something like that?

atavism:
@myleshorton Oh, sure. One reason I kept it the same: it sounds like we plan to support even more options (such as turning auto-detection on/off). Maybe it's significant enough a change it's worth coming up with a new design? It's the first screen a new user sees after all. I'll go ahead and make those changes for now though.

myleshorton:
Yeah I think a new design is probably a good idea. @carsonefv you want to maybe give it a look with the design team?

Let's not let that hold up beta3 I don't think though -- we can iterate on the design in parallel.

atavism:
@myleshorton Should be updated now.

myleshorton:
Thanks @atavism! For some reason I'm not seeing the welcome screen pop up even in an incognito window now. Should I be trying something different?

atavism:
@myleshorton Are you maybe using an earlier incognito session? You should open a new window. Alternatively, you can go to chrome://settings/cookies#cont, search for "127.0.0.1", delete our entry, and try opening the UI again. Also, are you seeing anything in the console?

src/github.com/getlantern/analytics/analytics.go
```diff
@@ -0,0 +1,92 @@
...
+// Makes a tracking request to Google Analytics
+func SendRequest(httpClient *http.Client, payload *Payload) (status bool, err error) {
+	if httpClient == nil {
+		log.Trace("Using default http.Client")
+		httpClient = defaultHttpClient
+	}
+
+	url := composeUrl(payload)
+
+	if _, err = http.NewRequest("GET", url, nil); err != nil {
```
oxtoacart: NewRequest doesn't actually send anything to the server.  Perhaps you want to use http.Get()?

src/github.com/getlantern/flashlight/settings/settings.go
```diff
@@ -39,3 +61,20 @@ func start(baseSettings *Settings) error {
...
 }
+
+func read() {
+	for msg := range service.In {
+		settings := (msg).(map[string]interface{})
+		config.Update(func(updated *config.Config) error {
+
+			autoReport := settings["autoReport"].(bool)
+			if autoReport {
+				analytics.Start()
```
oxtoacart: It would probably be better for analytics to have the typical Configure() method that gets called on config updates and then start/stop itself based on autoReport.  That way, if someone else changes the setting, it still works.

src/github.com/getlantern/flashlight/settings/settings.go
```diff
@@ -39,3 +61,20 @@ func start(baseSettings *Settings) error {
...
+			}
+
+			baseSettings.AutoReport = autoReport
```
oxtoacart, line 83: This needs to be protected by a mutex.

oxtoacart, line 83: See also line 58 where it needs to be protected by a mutex.

src/github.com/getlantern/lantern-ui/app/js/services.js
```diff
@@ -209,15 +196,32 @@ angular.module('app.services', [])
...
+              var gaObj = {
+                  clientId: '',
+                  clientVersion: '',
+                  language: '',
+                  screenColors: '',
+                  screenResolution: '',
+                  trackingId: '',
+                  viewPortSize: ''
+              };
+              for (var name in fields) {
```
oxtoacart, line 215: It would be great if we could capture the utma (visitor id) so that if the user is running the lantern UI in the same browser with which (s)he downloaded lantern, we can theoretically track them end-to-end.

src/github.com/getlantern/flashlight/analytics/analytics.go
```diff
@@ -0,0 +1,68 @@
...
+
+func Configure(autoReport *bool, newClient *http.Client) {
+	httpClient = newClient
+
+	if autoReport != nil && *autoReport {
+		err := Start()
+		if err != nil {
+			log.Errorf("Error starting analytics service: %q", err)
+		}
+	}
```
oxtoacart, line 55: Need to stop if autoReport is false.

src/github.com/getlantern/flashlight/settings/settings.go
oxtoacart: Oh, I just noticed that analytics already has a Configure() method.  Is this call redundant?

src/github.com/getlantern/flashlight/settings/settings.go
```diff
@@ -39,3 +61,20 @@ func start(baseSettings *Settings) error {
...
+			*updated.AutoReport = autoReport
+			return nil
+		})
```
oxtoacart, line 86: Do we also need to publish the updated settings to other clients?  Maybe that's actually best done in Configure()?

src/github.com/getlantern/lantern-ui/app/js/services.js
atavism, line 215: Good idea. How does the client id differ from the visitor id? I noticed that id is the same across clients. Can we just use it?

src/github.com/getlantern/analytics/analytics.go
atavism: Woah, thanks. I thought NewRequest actually sent the request. Should be fixed.

src/github.com/getlantern/flashlight/analytics/analytics.go
atavism, line 55: This implies the analytics service should also be shut down (if it's running) and removed from our UI services map. It doesn't look like there's any unregister functionality right now. I'll go ahead and add that?

src/github.com/getlantern/lantern-ui/app/js/services.js
oxtoacart, line 215: Ahh, interesting.  It looks like client ID effectively replaces utma - https://productforums.google.com/forum/#!topic/analytics/zfaZbC0E6pE

src/github.com/getlantern/flashlight/analytics/analytics.go
oxtoacart, line 55: Actually the requirement is just that we don't send the data to the cloud.  You could keep the service running and just discard whatever data the UI sends it.

src/github.com/getlantern/analytics/analytics.go
oxtoacart: Okay.  Please make sure to test that the data is actually getting into Google Analytics.  That can be a little annoying because it can take a little time for it to show up (at least, it used to), but eventually it should show up.

src/github.com/getlantern/analytics/analytics.go
```diff
@@ -0,0 +1,92 @@
...
+// Makes a tracking request to Google Analytics
+func SendRequest(httpClient *http.Client, payload *Payload) (status bool, err error) {
+	if httpClient == nil {
+		log.Trace("Using default http.Client")
+		httpClient = defaultHttpClient
+	}
+
+	url := composeUrl(payload)
+
+	if _, err = http.Get(url); err != nil {
```
oxtoacart: You need to capture the response and if there's no error, resp.Body.Close(), otherwise this can leak file descriptors.

oxtoacart: See the overview of http://golang.org/pkg/net/http/.

src/github.com/getlantern/flashlight/analytics/analytics.go
atavism, line 55: Right, but why do we need to keep the service running too? Why have an unused service present at all? Wouldn't it be better to close the connection and free up resources immediately?

oxtoacart, line 55: Yes, I think that would be nice if it's easy enough to do reliably.  That said, what resources are we talking about?  A socket and a few lines of code that run on every page view?  Actually, now that we're talking about it, it looks like we just track page view when the UI is opened, right?  Do we even need a websocket service for this, or would a REST endpoint suffice?

atavism, line 55: Mainly closing the socket connection and removing analytics from our services map. Just that would be cleaner when reporting is suddenly switched off. Yeah, even though we're just tracking the single page view right now, the reason for the websocket is we'll probably want to track additional user interaction at some point (different screens in the UI and eventually a browser extension), session duration, etc.

oxtoacart, line 55: I guess the salient question is whether or not the communication is always one way (i.e. client posting something to server).  If so, it's probably easiest just to make this a RESTful POST, and then you don't have to bother with starting/stopping services.  You just post from the client and, if reporting is enabled, the back-end posts that to the server.  If you want to get fancy, you can have the client eschew posting to the back-end when reporting is disabled, but I'd consider that icing on the cake.

myleshorton:
Stepping back on this for a second, guys, keep in mind we're pushing this to the backend for a few reasons:
1. We don't need to rely on the UI being open to record a session
2. We can also record analytics data from servers

#1 in particular will be extremely useful for tracking general Lantern use, not just installs. I also haven't reviewed the code in detail, so maybe we're already doing this. I just wanted to make sure we're not _only_ recording sessions on first startup.

atavism:
@myleshorton Right now, the request to GA happens whenever the UI is open. That gives us browser info (which we still want to collect, right?), operating system, and a unique identifier for clients. We don't have to rely on obtaining anything from the browser though. Should we generate random tracking and client IDs if there's no UI available?

oxtoacart:
I think this is a good example of where starting with requirements would help.  So, let's talk requirements.  What are the statistics that we want to track initially?

myleshorton:
> I think this is a good example of where starting with requirements would help.

Agreed! Let's actually just do that. 

At a high level, I think we do want all the info you mentioned, @atavism. In general I would say we do want information from the browser, ultimately including things like recording user interactions with settings and whatnot. The easiest way to do that might even be to just keep the frontend analytics, though. That would simplify the communication with the backend, and the js version of analytics also includes things like more intelligent backoff if things go awry.

You want to start a quick LEP we can iterate on @atavism?

atavism:
@myleshorton That's what I intended with the design though already: we gather all the default front-end analytics and send that to GA from the backend. Before we send that request, we can tack on whatever additional user information we want to. I still think it's fine to maintain a websocket connection for the service, in accordance with the other UI services we're already running, to track actual user interaction. Yeah, I can go ahead and assemble a list of all the parameters/interactions we're recording.

myleshorton:
Cool. Yeah I'm not totally against assembling it all on the backend. I do
definitely think we want to record backend sessions regardless though, both
for when the UI isn't running on system startup and for when we're just
running from a server.

We don't have to do anything with the server side yet but should at least
record all client-side sessions.

On Monday, March 23, 2015, atavism notifications@github.com wrote:

> @myleshorton https://github.com/myleshorton That's what I intended with
> the design though already: we gather all the default front-end analytics
> and send that to GA from the backend. Before we send that request, we can
> tack on whatever additional user information we want to. I still think it's
> fine to maintain a websocket connection for the service, in accordance with
> the other UI services we're already running, to track actual user
> interaction. Yeah, I can go ahead and assemble a list of all the
> parameters/interactions we're recording.
> 
> —
> https://github.com/getlantern/flashlight-build/pull/138#issuecomment-85198035
> .

## 

## 

President
https://www.getlantern.org

atavism:
@myleshorton Preliminary LEP available [here](https://github.com/getlantern/lantern-roadmap/blob/ssa/lep/DRAFT/0012_Server-Side_Analytics.md)

myleshorton:
Awesome! Looks good to me. The one area it seems like we should think more about while we're here is how to handle the server/give mode side. I wonder, for example, if there's any way to treat X request on a give mode server/peer like a normal visit almost as if we were a web page instead of a proxy. That way we could see traffic by region in a really intuitive way in the GA UI. What do you think?

atavism:
What we could do is set the location or hostname field to the server address and use a custom filter. If it's 127.0.0.1, we'll know the request originated from a client.

myleshorton:
Yeah that sounds good. You're saying because we're tunneling these requests
through localhost right?

On Wed, Mar 25, 2015 at 5:42 PM, atavism notifications@github.com wrote:

> What we could do is set the hostname field to the server address and use a
> custom filter. If it's 127.0.0.1, we'll know the request originated from a
> client.
> 
> —
> https://github.com/getlantern/flashlight-build/pull/138#issuecomment-86266725
> .

## 

## 

President
https://www.getlantern.org

atavism:
Yeah, basically.. To elaborate, hostname is an optional field "where your content is being hosted from" With the front-end, we were setting it arbitrarily to "lantern-ui". If we set that in give-mode, per request, to the actual server IP address, that'll give us an effective way to split traffic between both modes using a hostname filter. You can see this by going into GA "Geo -> Location" and adding "Behavior -> Hostname" as a secondary dimension.

myleshorton:
Sounds like a plan!

myleshorton:
Does this diff now also support just recording each session on the backend without a call from the frontend @atavism?

atavism:
@myleshorton The most recent diff actually sends an event to GA on start to track the session. Opening the UI fires a separate event now.

myleshorton:
Awesome. We can probably pull this in then. What do you think @oxtoacart are there remaining issues here?

atavism:
@myleshorton Found one issue actually starting/stopping the analytics service I need to resolve before it's ready for review.

atavism:
@myleshorton good to go now.

src/github.com/getlantern/flashlight/settings/settings.go
```diff
@@ -11,31 +17,70 @@ const (
...
+		settings := baseSettings
+		baseSettings.mutex.RUnlock()
+		return write(settings)
 	}
 
 	service, err = ui.Register(messageType, nil, helloFn)
+	go read()
 	return err
 }
+
```
myleshorton, line 71: Maybe just a comment/godoc here @atavism?

myleshorton:
OK -- the tests keep periodically failing, but it appears random. Working in all local tests, so merging.

PR #140, opened 2015-03-23
Loggly: don't treat colon in url as seperator

fffw:
Fixes https://github.com/getlantern/lantern/issues/2322

### DIFF
--- a/src/github.com/getlantern/flashlight/logging/logging.go
+++ b/src/github.com/getlantern/flashlight/logging/logging.go
@@ -179,24 +179,28 @@ func (w logglyErrorWriter) Write(b []byte) (int, error) {
 
 	// extract last 2 (at most) chunks of fullMessage to message, without prefix,
 	// so we can group logs with same reason in Loggly
-	parts := strings.Split(fullMessage, ":")
-	var message string
-	pl := len(parts)
-	switch pl {
-	case 1:
-		message = ""
-	case 2:
-		message = parts[1]
-	default:
-		message = parts[pl-2] + ":" + parts[pl-1]
+	lastColonPos := -1
+	colonsSeen := 0
+	for p := len(fullMessage) - 2; p >= 0; p-- {
+		if fullMessage[p] == ':' {
+			lastChar := fullMessage[p+1]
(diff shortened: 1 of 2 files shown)


src/github.com/getlantern/flashlight/logging/logging.go
```diff
@@ -179,24 +179,28 @@ func (w logglyErrorWriter) Write(b []byte) (int, error) {
...
-		message = parts[1]
-	default:
-		message = parts[pl-2] + ":" + parts[pl-1]
+	lastColonPos := -1
+	colonsSeen := 0
+	for p := len(fullMessage) - 2; p >= 0; p-- {
+		if fullMessage[p] == ':' {
+			lastChar := fullMessage[p+1]
+			// to prevent colon in "http://" and "x.x.x.x:80" be treated as seperator
+			if !(lastChar == '/' || lastChar >= '0' && lastChar <= '9') {
```
oxtoacart: I think that this is too specific.  If you do what I originally suggested, which is to take as many chunks as will fit into 100 characters, then that should largely take care of this problem.

fffw: That will leave too many message groups of which the only difference is the host to connect, like below

```
fronted: HTTP ERROR FOR MASQUERADE ritegamer.com: Head http://www.google.com/humans.txt: Unable to dial proxy: Unable to dial proxy: Unable to dial masquerade ritegamer.com: tlsdialer: DialWithDialer timed out\n
```

oxtoacart: True.  At the end of the day, once we're in the position of having to parse freeform text, it's kind of a losing proposition.  There's a separate ticket somewhere to start recording the line number on which a message was logged.  What if we just use that to group things?

fffw: Hm, the motive of truncating message is not to miss some item when grouping by it (https://github.com/getlantern/lantern/issues/2287), and we always have `fullMessage`. Grouping by location and by reason can be of different dimension. Anyway, it is insignificant I believe.

oxtoacart: Okay, sold.

PR #144, opened 2015-03-26
Check for GOARCH on win so Lantern can run closes getlantern/lantern#2334

myleshorton:


aranhoide:
I cheated and changed `"Cygwin"` to `"GNU/Linux"` to test this.

My GOARCH is `amd64` so if I run this with `./setenv.bash` then I get the error message.  So far so good.

But at least in Linux, for this to modify your environment you need to `source ./setenv.bash`.  If I do this, the `die` kills my terminal before I ever get a chance to see the error message.  If this is not the case in Cygwin, all is good.

I wonder whether you QA'd this by sourcing `setenv.bash` directly from the shell, or by using a script that sources it (like `crosscompile.bash` or `run.bash`)?

myleshorton:
Good catch @aranhoide -- yeah I was testing with calling from another script. Just added a check for that before calling exit. How's this work for you?

aranhoide:
This version doesn't kill my terminal, but it doesn't abort the script either, so it will go on to set the GOPATH.  Probably not what we want in that case.

myleshorton:
So this is a little bit squirrelly, but basically the only reasons I put that go architecture check in setenv.bash at all is that:
1. It's the one script that everyone needs to call to build flashlight correctly
2. We're already checking for cygwin there and only there

Otherwise it's kind of the wrong place because setenv.bash doesn't actually do the build. We only really want it to abort when it is a part of a build, which is currently when it's called from other scripts.

In terms of setting the GOPATH, it actually sets the GOPATH correctly no matter what, so the fact that it doesn't abort when sourced directly in the shell is actually OK. We mostly want to abort when we're actually building and/or running, as it's only those cases where things will actually go awry.

So I _think_ the current version is as close to what we want as I can think of, but I'm totally open to suggestions. Another option would be to kill setenv.bash entirely and add that functionality to run.bash and crosscompile.bash or just document that people should run those and no source setenv.bash directly.

aranhoide:
I just wanted to make sure that you were aware of the issue.  I don't think it's a big deal either, esp. since nobody does development work on Cygwin; I expect we'll only ever do releases from that one.

aranhoide:
For the record, I think setenv.bash is worth keeping to be able to `go build` particular projects while in development.

myleshorton:
Makes sense thanks!

PR #145, opened 2015-03-28
NL chained proxies

aranhoide:
I have launched 100 old-style fallback (chained) proxies in the Amsterdam 3 DO datacenter.  This change replaces `fallback.json` to point at the new servers **instead** of the old ones, which are left to serve users of legacy Lantern versions that can't take advantage of the `lantern.yaml` chained proxy list.  It also regenerates files (by just running `genconfig/genconfig.bash`).  I haven't deployed the resulting `cloud.yaml` anywhere.  You can either do it on merge or let me know and I'll do it.

I updated the README to explain how to generate a new `fallback.json`.  The README was also out of date with recent additions to genconfig functionality, so I added a mention to those.


myleshorton:
Awesome!!! Thanks for knocking this out. LGTM.

myleshorton:
Just fiddling with this some more @aranhoide. It's odd so many of the fallbacks fail their checks and don't make it into cloud.yaml as a result. See:

```
genconfig (valencia *)]$ grep fallback- cloud.yaml | wc -l
69
```

There are a lot of:

```
genconfig: Skipping fallback 178.62.213.97 because dialing Google failed: Unable to dial server: EOF
```

when I run genconfig.bash

myleshorton:
It looks like this was a problem with the current generated cloud.2.0.0-beta3.yaml too -- that only has 75 fallbacks in it.

Very odd that so many brand new fallbacks would fail like that. I suspect there's something amiss on the client side.

myleshorton:
I went ahead and uploaded lantern-config/cloud.2.0.0-beta4-03-28-15.yaml.gz

PR #150, opened 2015-03-31
nonStopWriter never fails 

fffw:
Resolves getlantern/lantern#2354

Writing to stdout and stderr will fail if `-Hwindowsgui` supplied, `timestamped` wil stop to write log message if it sees any error while writing timestamp. this PR correct it.


myleshorton:
LGTM.

PR #151, opened 2015-03-31
Merging release script tweaks to valencia

myleshorton:


myleshorton:
This makes it so that we always first release a special QA build that @sheena3 can take a look at before we release it to beta.

myleshorton:
Peerscanner test failing -- logged.

PR #152, opened 2015-04-01
update fallback (chained) proxy data

aranhoide:
The certs for many chained proxies have changed to fix:

https://github.com/getlantern/lantern/issues/2341

(When you merge this either upload the corresponding cloud.yaml or let me know so I do it.)


myleshorton:
Awesome! Do you have a sense of if/when we've broken cloud.yaml compatibility in the past? I.e. How far back can we replace old cloud.yamls with this one?

aranhoide:
@myleshorton : Sorry, I didn't know from the top of my head and checking this fell through the cracks..

I diffed the default cloud.yaml from 2.0.0-beta1 to 2.0.0-beta4 and the only difference, besides the addition of the chained proxies and the IPs of the masquerades, was some `autoreport: true` which was introduced recently but doesn't seem to break anything in -beta1.  I run -beta1 a bit with -beta4's config.yaml and I couldn't spot any problem.  Traffic went through the old-style fallbacks in NY, as expected [1].

[1] Note that for some reason 2.0.0-beta4 comes configured to fetch its config from -beta3:

https://s3.amazonaws.com/lantern_config/cloud.2.0.0-beta3.yaml.gz .

aranhoide:
tl;dr: it should be safe to feed this new config to all clients in the 2.x line (that is, all pure Golang clients).

myleshorton:
OK cool thanks for checking! 

> [1] Note that for some reason 2.0.0-beta4 comes configured to fetch its config from -beta3:
> 
> https://s3.amazonaws.com/lantern_config/cloud.2.0.0-beta3.yaml.gz .

The numbering here just reflects the Lantern version at the time a change was made. I.e. we don't release a new cloud.yaml for every Lantern version but rather just name it according to the current version whenever it does change. 

Since these are all betas, here's what I think we should do:
1. Upload the new cloud.yaml
2. Point it at https://s3.amazonaws.com/lantern_config/cloud.2.0.0.yaml.gz 
3. Continually point to that version from here on and make sure to maintain backwards compatibility for all 2.x versions moving forward.

That sound OK?

aranhoide:
Sounds good.  Initially the new cloud.yaml will need to be uploaded to all the old cloud.yaml URLs so old clients will pick it up, but that'll be the last titme we need to touch those URLs.

myleshorton:
Sounds great. Do you mind just executing that last step? Will be cool to get this in there, although I guess we also have Tokyo coming online super soon too hopefully.

aranhoide:
Yes, will do.

And yes, we can have Tokyo fallbacks and flashlight proxies tomorrow if we want.  Doing the Vultr salt-cloud module is a nice to have in the long run but not a prerequisite.

The problem with just adding the Tokyo servers to the current cloud.yaml is that AFAIK the client still doesn't know how to pick the fastest one, so (e.g.) Iran users would be proxying through Tokyo about half the time.  Even if we fix that for new clients, we can't retroactively fix it for clients already in the field.  So as soon as we publish a cloud.yaml that has Tokyo servers, we'll probably want to diverge again.

myleshorton:
Oh yeah good point. We definitely have to add that to the client side.
Given that the vast majority of old users are in China, we could also just
point all the old ones to Tokyo and diverge with the new.

We discussed on the call today shooting for having beta 5 in QA by the end
of the week, so let's just try to get all of this in by end of day Wed if
you can so we can make that happen.

That sound good?

On Mon, Apr 6, 2015 at 10:36 PM, aranhoide notifications@github.com wrote:

> Yes, will do.
> 
> And yes, we can have Tokyo fallbacks and flashlight proxies tomorrow if we
> want. Doing the Vultr salt-cloud module is a nice to have in the long run
> but not a prerequisite.
> 
> The problem with just adding the Tokyo servers to the current cloud.yaml
> is that AFAIK the client still doesn't know how to pick the fastest one, so
> (e.g.) Iran users would be proxying through Tokyo about half the time. Even
> if we fix that for new clients, we can't retroactively fix it for clients
> already in the field. So as soon as we publish a cloud.yaml that has Tokyo
> servers, we'll probably want to diverge again.
> 
> —
> https://github.com/getlantern/flashlight-build/pull/152#issuecomment-90370059
> .

## 

## 

President
https://www.getlantern.org

aranhoide:
Oh, I thought the majority of Lantern Valencia users were from Iran?  If they're mostly in China then yes, I'd better deploy the Tokyo servers and publish a cloud.yaml that points to them instead.  Not much value in publishing the NL cloud.yaml for a day or so.

myleshorton:
No they are mostly from Iran, but almost none prior to that were.

aranhoide:
For 1.x Java-based Lantern clients, most traffic goes to the single fallback that they get from the controller, not from the cloud.yaml.  So changing the cloud.yaml would have very limited impact either way.  I would just leave the cloud.yaml(s) used by 1.x clients alone.

For 2.x golang-based Lantern clients already deployed, enabling the NL chained servers would help with performance in any case (since Amsterdam works better than New York for practically everyone using 2.0 currently) and further cut down on our usage of domain fronting, so I think it's worth doing.

aranhoide:
Wait, did you mean point the 1.x cloud.yaml to Tokyo?  Since that one doesn't need chained servers, a quick way to do that would be just editing the frontedservers: entries to point to `jp.fallbacks.getiantem.org` (once we actually have flashlight servers running there, of course).

aranhoide:
Whatever we do about the 1.x line, I think it's clear that this is worth deploying to 2.0, so I'll merge it and upload it.

PR #153, opened 2015-04-01
Add headless build option

fffw:
set `HEADLESS_BUILD=true` before running 'crosscompile.bash' and 'linuxcompile.bash', so it can be built and run at platforms not have systray support libraries.

Resolves https://github.com/getlantern/lantern/issues/2347

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,6 +29,9 @@ define build-tags
 		sed s/'packageVersion.*'/'packageVersion = "'$$VERSION'"'/ src/github.com/getlantern/flashlight/autoupdate.go | sed s/'!prod'/'prod'/ > src/github.com/getlantern/flashlight/autoupdate-prod.go; \
 	else \
 		echo "** VERSION was not set, using git revision instead ($(GIT_REVISION)). This is OK while in development."; \
+	fi && \
+	if [[ -z "$$HEADLESS" ]]; then \
+		BUILD_TAGS="$$BUILD_TAGS headless"; \
 	fi
 endef
 
--- a/Makefile
+++ b/Makefile
@@ -32,7 +32,9 @@ define build-tags
 	fi && \
 	if [[ ! -z "$$HEADLESS" ]]; then \
 		BUILD_TAGS="$$BUILD_TAGS headless"; \
-	fi
+	fi && \
+	BUILD_TAGS=$$(echo $$BUILD_TAGS | xargs) && \
+	echo "Build tags: $$BUILD_TAGS"
 endef
 
 define docker-up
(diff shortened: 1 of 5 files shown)


myleshorton:
@aranhoide I know you had some thoughts on this -- you mind doing the honors on the review?

fffw:
As https://github.com/getlantern/flashlight-build/pull/154 was merged, should update to use Makefile accordingly before merge.

fffw:
@xiam This is the PR to add headless option.

I'm not good at Makefile, do you have any idea what causes this error?

>  ~/workspace/getlantern/flashlight-build (issue2347) $ make linux
> Generating assets...
> make: *** No rule to make target 'docker-genassets'.  Stop.
> make: *** [genassets] Error 2

fffw:
Oh, I don't have anything under `/flashlight-build`! Will check.

fffw:
@xiam Below command should list all files under host directory right? But I only got `total 0`, really odd.

```
make docker
docker run -v /Users/fffw/workspace/getlantern/flashlight-build:/flashlight-build -t flashlight-builder /usr/bin/ls -l
```

xiam:
Odd. Are you on Mac or Linux? A ls / gives you nothing too?

fffw:
Mac with latest boot2docker. Running `/usr/bin/ls /` is fine. The volume seems simply not get mapped.


BR,Merlin

fffw:
@xiam I updated Makefile. You can try the changes on your docker setup to build headless version if you like.

Makefile
```diff
@@ -29,6 +29,9 @@ define build-tags
 		sed s/'packageVersion.*'/'packageVersion = "'$$VERSION'"'/ src/github.com/getlantern/flashlight/autoupdate.go | sed s/'!prod'/'prod'/ > src/github.com/getlantern/flashlight/autoupdate-prod.go; \
 	else \
 		echo "** VERSION was not set, using git revision instead ($(GIT_REVISION)). This is OK while in development."; \
+	fi && \
+	if [[ -z "$$HEADLESS" ]]; then \
```
xiam: I think this should be `! -z`, since we want to check when $$HEADLESS is not empty.

fffw: Ahh yes. I don't have a workable build environment to test Makefile so feel
free to modify it. I'm also trying to set up docker on Ubuntu VM, hopefully
I can test it tomorrow morning.

xiam: OK! just making sure, otherwise looks fine here. I used this to build the headless Lantern I needed to check sonar.

myleshorton:
I can't remember offhand @aranhoide, but I thought you had some issue with building separate headless binaries. Am I misremembering that?

myleshorton:
I'm just catching up on the discussion at #2347. Looks like we're all agreed with just doing a separate headless build, so please ignore my above comment.

aranhoide:
@xiam : How did you pass the `HEADLESS` environment variable?  If it is just set in the environment of the caller, or if I just call `HEADLESS=true make linux-amd64` then it's not picked up and the resulting binary still depends on libappindicator3.  OTOH, if I remove one of the `$`s from `$$HEADLESS` in the Makefile, that works.  This seems to match what I read in

https://www.gnu.org/software/make/manual/html_node/Environment.html

> Variables in make can come from the environment in which make is run. Every environment variable that make sees when it starts up is transformed into a make variable with the same name and value. 

Since it's a make variable it makes sense that it gets accessed with a single `$`, right?

Are you sure that the binary you got doesn't work because you happened to have libappindicator in the machine you used to check sonar?

xiam:
Hello @aranhoide,

I'm pretty sure it works with passing the HEADLESS env, I used that to test
a PR last week. No risk of me having libappindicator as I tested on a clean
vm. Let me check again, maybe I missed something. Did you see my last
commit?

Just for the record,

Currently, HEADLESS is an environment variable, not a Make one, so we must
escape it for the target to see it, otherwise it will be expanded by Make
before having a chance to be run by the target.

aranhoide:
@xiam : you run that in OS X, right?  Maybe that's a difference with the make in Linux?  Again, see the docs above

> Every environment variable that make sees when it starts up is transformed into a make variable

Treating it as an environment variable doesn't work for me in Ubuntu 14.10.  Treating it as a make variable (that is, only one `$`), does work.

xiam:
@aranhoide Oh, you're right! I see now what you mean, I was confusing Make variables with macros. Sorry about that. But it looks like they still need to be escaped with the double $. I tested echoing the $HEADLESS inside the build-tags and I got a "EADLESS" string (it treats $H as a *Make variable*, which is not defined), that's why it was working to you, but probable `HEADLESS="" make linux-amd64` would produce the same headless build, even when the HEADLESS env is empty.

I also found the problem, the HEADLESS value was not being passed to the docker-linux-amd64 target, I don't know how did I managed to compile the thing last week, I probably exported an special env var at some point and that's why my build ended up being without the libapp dependency.

Please try my newest commit, libappindicator should not be included anywhere:

```
strings lantern_linux_amd64 | grep appindicator
```

Note that this branch still dies with the following message if the translations are not set:

```
flashlight: Not found any translations, locale not set
```

That was [already fixed in valencia](https://github.com/getlantern/flashlight-build/commit/b7b0650481cefa673c16e05fcb87ce73a7e1b437).

A workaround for this branch would be setting a VERSION:

```
VERSION=2.0.0-beta5 HEADLESS=true make linux-amd64 
```

That would create a binary ready for production, with translations included.

aranhoide:
@xiam you don't seem to have pushed that latest commit, have you?  At least not to the issue2347 branch...

Makefile
```diff
@@ -29,7 +29,12 @@ define build-tags
 		sed s/'packageVersion.*'/'packageVersion = "'$$VERSION'"'/ src/github.com/getlantern/flashlight/autoupdate.go | sed s/'!prod'/'prod'/ > src/github.com/getlantern/flashlight/autoupdate-prod.go; \
 	else \
 		echo "** VERSION was not set, using git revision instead ($(GIT_REVISION)). This is OK while in development."; \
-	fi
+	fi && \
+	if [[ ! -z "$$HEADLESS" ]]; then \
+		BUILD_TAGS="$$BUILD_TAGS headless"; \
+	fi && \
+	BUILD_TAGS=$$(echo $$BUILD_TAGS | xargs) && \
```
xiam: This is for trimming BUILD_TAGS

xiam:
@aranhoide Sorry, I mistakenly pushed it to a new issue-2347 branch, with a dash. I just fixed that.

aranhoide:
@xiam that did it, thanks!  Although the version in this branch won't work in server mode, that is for an unrelated reason.  I tested that a version built by merging valencia into this branch does work in a server without the libappindicator library.

PR #154, opened 2015-04-01
Dockerfile/Makefile and instructions for cross compiling Lantern (including CGO features) and generating packages

xiam:
This PR replaces Lantern's building and packing scripts with a single `Makefile`. The main benefit GNU Make provides over simple bash scripts is the ability to define dependencies, catch errors and connect different tasks into a single one. Some tasks within this Makefile are deferred to a docker image, this docker image is built according to instructions defined in the included `Dockerfile`.

In order to clean installation dependencies, the `installer-resources` directory was also introduced with this PR.

Note that some other scripts (such as `updateicons.bash` or `createrelease.rb`) were not included into this Makefile since the proposal is focused on building and packing only.


myleshorton:
This looks awesome @xiam! Merging.

PR #155, opened 2015-04-02
Providing support for bzip2 compressed assets

xiam:
This PR provides support for reading bzip2 compressed assets on the update server. The Makefile was also modified in order to generate such assets.

### DIFF
--- a/src/github.com/getlantern/autoupdate-server/server/asset.go
+++ b/src/github.com/getlantern/autoupdate-server/server/asset.go
@@ -49,9 +49,7 @@ func downloadAsset(uri string) (localfile string, err error) {
 			return "", fmt.Errorf("Expecting 200 OK, got: %s", res.Status)
 		}
 
-		if res.Body != nil {
-			defer res.Body.Close()
-		}
+		defer res.Body.Close()
 
 		var fp *os.File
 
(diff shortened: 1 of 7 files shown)


xiam:
Hello @atavism, you helped me reviewing the auto-update PR, could you help me taking a look at this too?

Thanks,

atavism:
@xiam Sure!

src/github.com/getlantern/autoupdate-server/server/asset.go
```diff
@@ -46,6 +49,10 @@ func downloadAsset(uri string) (localfile string, err error) {
 			return "", fmt.Errorf("Expecting 200 OK, got: %s", res.Status)
 		}
 
+		if res.Body != nil {
+			defer res.Body.Close()
+		}
+
```
atavism, line 53: If http.Get doesn't return an error, do we need to check the response body for nil here?

xiam, line 53: > When err is nil, resp always contains a non-nil resp.Body. Caller should close resp.Body when done reading from it.

Right @atavism! this is not really required. Fixed.

atavism:
LGTM

PR #156, opened 2015-04-03
Logging error instead of exiting, a missing translation should not be a ...

xiam:
Closes #2360.


fffw:
LGTM. It resolves https://github.com/getlantern/lantern/issues/2360

PR #157, opened 2015-04-05
update dockerfile and makefile

fffw:


Dockerfile
```diff
@@ -75,6 +75,6 @@ ENV SECRETS /secrets
 RUN mkdir -p $WORKDIR
 RUN mkdir -p $SECRETS
 
-VOLUME [ $WORKDIR ]
+VOLUME [ "$WORKDIR" ]
```
fffw: extra directory '[' and ']' will be created without the quotation mark, see https://docs.docker.com/reference/builder/#volume

fffw:
@xiam just a small fix.

myleshorton:
LGTM!

PR #158, opened 2015-04-05
Integrate VirusTotal checking as part of release process

fffw:
Resolves https://github.com/getlantern/lantern/issues/2353

Added check-virus-windows target, and let package-windows depends on it.


myleshorton:
Hey thanks @fffw I totally forgot to update the status of my playing around in a very similar way to what you've done here. The big problem is that virustotal processes API requests last. This is from their docs at https://www.virustotal.com/en/documentation/public-api/:

> Keep in mind that files sent using the API have the lowest scanning priority, depending on VirusTotal's load, it may take several hours before the file is scanned, so query the report at regular intervals until the result shows up and do not keep sending the file once and over again.

So we can certainly auto-submit it via the API, but I think Sunshine will have to check it in QA manually probably. The only other approach I can think of is to figure out something else that automatically checks the VT status of the latest binaries, like maybe Travis or even something that just checks it with ```go test``` locally? Not sure.

fffw:
My original idea is that we don't need to check the virus report often, maybe just before release each new version as a safety net. The script shows me report in 1 minute or even 20 seconds during my test run, so I don't expect it to be very time consuming. We can ignore the time required if we employ a build server later.

myleshorton:
That sounds great @fffw. If it's really that quick in practice, we can totally just have the release scripts block on it probably. I thought it was much longer in my testing, but if it's that short most of the time let's just keep it in the release scripts.

myleshorton:
What's the thinking behind checking the executable versus the installer? It seems like the installer will cover the executable automatically while also capturing anything that might get flagged in the installer itself (probably nothing in practice, but we might add more to the installer later).

fffw:
Maybe we should check it twice? One for the executable and one for the installer. The results are not quite the same, [installer](https://www.virustotal.com/en/file/e3ba8db0e2f40ebd3fc2688f32817a92d84b523538dbd9892623e5e865ab257c/analysis/1427761541/) versus [executable](https://www.virustotal.com/en/file/687460d74363992af8db87c4eab878d9ab7c1287f6504592e149904693f46c7c/analysis/).

myleshorton:
Makes sense to me.

On Sun, Apr 5, 2015 at 3:37 PM, fffw <notifications@github.com> wrote:

> Maybe we should check it twice? One for the executable and one for the
> installer. The results are not quite the same, installer
> <https://www.virustotal.com/en/file/e3ba8db0e2f40ebd3fc2688f32817a92d84b523538dbd9892623e5e865ab257c/analysis/1427761541/>
> versus executable
> <https://www.virustotal.com/en/file/687460d74363992af8db87c4eab878d9ab7c1287f6504592e149904693f46c7c/analysis/>
> .
>
> —
> <https://github.com/getlantern/flashlight-build/pull/158#issuecomment-89866627>
> .
>



-- 
--
President
https://www.getlantern.org

fffw:
Revoke this PR, as `virustotal` can be extremely slow even for files uploaded through web page.

PR #159, opened 2015-04-06
remove superfluous escape on '

aranhoide:
It doesn't seem necessary at least for GNU Make, and the backslash
will show up in the Ubuntu information panel when installing.

I'm sending this in a PR so hopefully someone tests this in OS X.  I haven't set up my new laptop yet.


xiam:
Right! I missed that one. Thanks for noticing.

PR #160, opened 2015-04-07
close connection before close response body in enproxy

fffw:
To resolve getlantern/lantern#2317

### DIFF
--- a/src/github.com/getlantern/enproxy/conn_impl_reads.go
+++ b/src/github.com/getlantern/enproxy/conn_impl_reads.go
@@ -17,11 +17,14 @@ func (c *conn) processReads() {
 
 	defer func() {
 		increment(&readingFinishing)
-		if resp != nil {
-			resp.Body.Close()
-		}
+		// be sure to close connection before close response body,
+		// or it will continuously receives data until hit EOF,
+		// which is a waste of bandwidth.
 		if proxyConn != nil {
 			proxyConn.conn.Close()
+		}
+		if resp != nil {
+			resp.Body.Close()
 		}
 		c.doneReadingCh <- true
 		decrement(&readingFinishing)


fffw:
I was trying to add unit test but seems not make much sense, can we skip that as this bug doesn't impact the functionality?

myleshorton:
Great catch @fffw! I wonder if this is also an issue in other places where we close the HTTP response body? We certainly do that all over the place. Is it always necessary to close the connection, or is there something special about that underlying enproxy connection? Perhaps it doesn't implement close properly itself (under the assumption that Body.Close() actually closes the underlying connection at least in some cases)?

Either way, great catch. Merging!

myleshorton:
Actually, @fffw, is it always the case that we want to close the connection here? I would think Body.Close would only actually close the connection if Response.Close is true -- see https://golang.org/src/net/http/response.go

In particular, if it's an HTTP 1.1 connection, we don't want to close the connection unless the Connection: close header is present, and that use case is very common.

It seems like we want something a bit different, i.e. to explicitly close the connection if we've lost the connection from the browser. 

I'm actually going to revert that merge until we get some clarity on that.

fffw:
@myleshorton It's not very clear for me the point here. This defer block only executes when error encountered, upstream EOF hit, or `readRequestsCh` is closed (explicitly calls enproxy connection's `Close()`). The connection should be closed anyway.

The change here is just closing proxy connection before closing `resp.Body`, to avoid wasting time and bandwidth on receiving the large chunks of response we don't need.

If we want to be aware of http header, that would be a separate feature ticket. We can't inspect https header anyway.

myleshorton:
OK that makes sense @fffw. So the issue before was that the resp.Body.Close wouldn't always close the actual connection, is that right?

myleshorton:
Can you open a new PR for it? Sorry for the inconvenience!

fffw:
@myleshorton Sorry I didn't describe it clearly. The issue is that `resp.Body.Close` hangs for a long time until all response data has received for large downloads.  There seems no other way to interrupt it except closing the connection. Swap the order so that it can return immediately.

myleshorton:
Interesting @fffw. I think your fix sounds correct. It actually sounds like it could be a bug in ```resp.Body.Close``` but let's just implement your fix for now.

PR #161, opened 2015-04-07
fix nil pointer in server mode

fffw:
I'm not sure if analytics should be configured at all in server mode, but pass a `nil` will cause a panic in [analytics](https://github.com/getlantern/flashlight-build/blob/valencia/src/github.com/getlantern/flashlight/analytics/analytics.go#L31)

Fixes https://github.com/getlantern/lantern/issues/2364.


aranhoide:
LGTM, I tested this to fix the issue.

aranhoide:
Thanks!

PR #165, opened 2015-04-09
Check subdomains in detour whitelist

fffw:
Our baked whitelist contains only primary domains, should check recursively to see if a site is in whitelist.

Resolves https://github.com/getlantern/lantern/issues/2371.

### DIFF
--- a/src/github.com/getlantern/detour/whitelist.go
+++ b/src/github.com/getlantern/detour/whitelist.go
@@ -1,13 +1,12 @@
 package detour
 
 import (
+	"strings"
 	"sync"
-	"time"
 )
 
 type wlEntry struct {
 	permanent bool
-	addTime   time.Time
 }
 
 var (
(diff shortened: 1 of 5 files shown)


fffw:
@aranhoide have interest to take a look?

src/github.com/getlantern/detour/whitelist.go
```diff
@@ -15,47 +14,56 @@ var (
 	whitelist   = make(map[string]wlEntry)
 )
 
-func InitWhitelist(wl map[string]time.Time) {
```
fffw: After changed `AddToWl` and `RemoveFromWl` to public, `InitWhitelist` no longer required.

src/github.com/getlantern/detour/whitelist.go
```diff
@@ -15,47 +14,56 @@ var (
...
 
-func DumpWhitelist() (wl map[string]time.Time) {
-	wl = make(map[string]time.Time)
+func RemoveFromWl(addr string) {
+	muWhitelist.Lock()
+	defer muWhitelist.Unlock()
+	delete(whitelist, addr)
+}
+
+func DumpWhitelist() (wl []string) {
```
fffw: `DumpWhitelist` is also unused. But it may has potential usage perhaps, so I just leave it here, but I changed the signature.

src/github.com/getlantern/detour/whitelist.go
```diff
@@ -15,47 +14,56 @@ var (
...
 	return
 }
 
-func whitelisted(addr string) bool {
+func whitelisted(addr string) (in bool) {
 	muWhitelist.RLock()
 	defer muWhitelist.RUnlock()
-	_, in := whitelist[addr]
-	return in
+	for ; addr != ""; addr = getParentDomain(addr) {
```
fffw: Recursively check if its primary domain is in whitelist.

src/github.com/getlantern/detour/whitelist_test.go
```diff
@@ -0,0 +1,20 @@
...
+
+import (
+	"testing"
+
+	"github.com/getlantern/testify/assert"
+)
+
+func TestCheckSubdomain(t *testing.T) {
+	AddToWl("facebook.com:80", true)
+	assert.True(t, whitelisted("www.facebook.com:80"), "should match subdomain")
```
myleshorton: Just add another test case for multiple subdomain?

src/github.com/getlantern/detour/whitelist.go
myleshorton: Not a biggie, but we could slightly optimize this through starting with checking the root domain and then adding subdomains since there are very few sites in the whitelist that include subdomains.

In practice, however, it's probably not a performance issue at all, and it would be slightly more complicated, so let's stick with what you've got!

myleshorton:
I'm just going to merge this @fffw -- looking good.

PR #170, opened 2015-04-10
add capability for country-specific cloud configs

aranhoide:
I have uploaded configs for 'cn' (goes to Tokyo) and 'default' (to Amsterdam).

Note that I couldn't run the unit tests for flashlight because bare flashlight won't build anymore.  I don't suppose this problem has anything to do with this PR:

```
# github.com/getlantern/flashlight
/usr/bin/ld: i386 architecture of input file `/var/tmp/go-link-AEmgl0/000000.o' is incompatible with i386:x86-64 output
collect2: error: ld returned 1 exit status
/home/aranhoide/src/gonative1.4.2/go/pkg/tool/linux_amd64/6l: running gcc failed: unsuccessful exit status 0x100
```

### DIFF
--- a/src/github.com/getlantern/flashlight/config/config.go
+++ b/src/github.com/getlantern/flashlight/config/config.go
@@ -31,6 +31,9 @@ import (
 const (
 	CloudConfigPollInterval = 1 * time.Minute
 
+	// In seconds
+	waitForLocationTimeout = 20
+
 	cloudflare         = "cloudflare"
 	etag               = "ETag"
 	ifNoneMatch        = "If-None-Match"
@@ -359,18 +362,26 @@ func (cfg Config) fetchCloudConfigForAddr(proxyAddr string) ([]byte, error) {
 	if err != nil {
 		return nil, fmt.Errorf("Unable to initialize HTTP client: %s", err)
 	}
-	country := strings.ToLower(globals.GetCountry())
-	if country != "" {
-		ret, err := cfg.fetchCloudConfigForCountry(client, country)
-		// We could check specifically for a 404, but S3 actually returns a 403 when
-		// a resource is not available.  I thought we'd better lean on the side of
-		// robustness by avoiding hardcoding an S3-ism here, than to try and save an
-		// extra request every now and then.
-		if err == nil {
-			return ret, err
-		} else {
-			log.Debugf("Couldn't fetch cloud config for country '%s'; trying the default one", country)
+	// Try and wait to get geolocated, up to a point.  Waiting indefinitely
+	// might prevent us from ever getting geolocated if our current
+	// configuration is preventing domain fronting from working.
+	for i := 0; i < waitForLocationTimeout; i++ {
+		country := strings.ToLower(globals.GetCountry())
+		if country != "" && country != "xx" {
+			ret, err := cfg.fetchCloudConfigForCountry(client, country)
+			// We could check specifically for a 404, but S3 actually returns a 403 when
+			// a resource is not available.  I thought we'd better lean on the side of
+			// robustness by avoiding hardcoding an S3-ism here, than to try and save an
+			// extra request every now and then.
+			if err == nil {
+				return ret, err
+			} else {
+				log.Debugf("Couldn't fetch cloud config for country '%s'; trying the default one", country)
+			}
+			break
 		}
+		log.Debugf("Waiting for location...")
+		time.Sleep(1 * time.Second)
 	}
 	return cfg.fetchCloudConfigForCountry(client, "default")
 }
--- a/src/github.com/getlantern/flashlight/config/config.go
+++ b/src/github.com/getlantern/flashlight/config/config.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"path/filepath"
 	"sort"
+	"strings"
 	"time"
 
 	"github.com/getlantern/appdir"
@@ -30,15 +31,16 @@ import (
 const (
 	CloudConfigPollInterval = 1 * time.Minute
 
-	cloudflare  = "cloudflare"
-	etag        = "ETag"
-	ifNoneMatch = "If-None-Match"
+	cloudflare         = "cloudflare"
+	etag               = "ETag"
+	ifNoneMatch        = "If-None-Match"
+	countryPlaceholder = "${COUNTRY}"
 )
 
 var (
 	log                 = golog.LoggerFor("flashlight.config")
 	m                   *yamlconf.Manager
-	lastCloudConfigETag = ""
+	lastCloudConfigETag = map[string]string{}
 )
 
 type Config struct {
@@ -209,7 +211,7 @@ func (cfg *Config) ApplyDefaults() {
 	}
 
 	if cfg.CloudConfig == "" {
-		cfg.CloudConfig = "https://s3.amazonaws.com/lantern_config/cloud.2.0.0-nl.yaml.gz"
+		cfg.CloudConfig = fmt.Sprintf("https://s3.amazonaws.com/lantern_config/cloud.%v.yaml.gz", countryPlaceholder)
 	}
 
 	// Default country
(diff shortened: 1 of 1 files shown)


aranhoide:
I recommend that you merge [this one](https://github.com/getlantern/flashlight-build/pull/171) too before cutting the next beta, or users in China will still proxy through Amsterdam about half the time (since their clients come with the Amsterdam servers preconfigured).

src/github.com/getlantern/flashlight/config/config.go
```diff
@@ -357,19 +359,36 @@ func (cfg Config) doFetchCloudConfig(proxyAddr string) ([]byte, error) {
...
+		}
+	}
+	return cfg.fetchCloudConfigForCountry(client, "default")
+}
 
-	log.Debugf("Checking for cloud configuration at: %s", cfg.CloudConfig)
-	req, err := http.NewRequest("GET", cfg.CloudConfig, nil)
+func (cfg Config) fetchCloudConfigForCountry(client *http.Client, country string) ([]byte, error) {
+	url := strings.Replace(cfg.CloudConfig, countryPlaceholder, country, 1)
+	log.Debugf("Checking for cloud configuration at: %s", url)
```
atavism, line 391: Just to confirm, in server-mode, the way to use a country-specific cloud config is via the `-cloudconfig http://s3.amazonaws.com/lantern_config/cloud.$\{COUNTRY\}.yaml.gz` command line designation. Is that correct? This substitutes whatever ISO country code `getCountry()` in `globals.go` returns; for our Tokyo servers, that's always expected to return "cn" (case-sensitive) to fetch `http://s3.amazonaws.com/lantern_config/cloud.cn.yaml.gz`?

src/github.com/getlantern/flashlight/config/config.go
```diff
@@ -357,19 +359,36 @@ func (cfg Config) doFetchCloudConfig(proxyAddr string) ([]byte, error) {
 	if err != nil {
 		return nil, fmt.Errorf("Unable to initialize HTTP client: %s", err)
 	}
+	country := strings.ToLower(globals.GetCountry())
```
myleshorton: This relies on globals.go SetLocation having already been called at this point right? Otherwise GetCountry will return nil.

src/github.com/getlantern/flashlight/config/config.go
```diff
@@ -209,7 +211,7 @@ func (cfg *Config) ApplyDefaults() {
 	}
 
 	if cfg.CloudConfig == "" {
-		cfg.CloudConfig = "https://s3.amazonaws.com/lantern_config/cloud.2.0.0-nl.yaml.gz"
+		cfg.CloudConfig = fmt.Sprintf("https://s3.amazonaws.com/lantern_config/cloud.%v.yaml.gz", countryPlaceholder)
```
myleshorton, line 217: Can we actually change this name to something like CloudConfigUrl? It's a bit confusing otherwise.

src/github.com/getlantern/flashlight/config/config.go
```diff
@@ -357,19 +359,36 @@ func (cfg Config) doFetchCloudConfig(proxyAddr string) ([]byte, error) {
...
 	}
+	country := strings.ToLower(globals.GetCountry())
+	if country != "" {
+		ret, err := cfg.fetchCloudConfigForCountry(client, country)
+		// We could check specifically for a 404, but S3 actually returns a 403 when
+		// a resource is not available.  I thought we'd better lean on the side of
+		// robustness by avoiding hardcoding an S3-ism here, than to try and save an
+		// extra request every now and then.
+		if err == nil {
+			return ret, err
```
myleshorton: If there's a non-200-level response here, I believe this won't return an error. So I think we need to check for a non-200 response here.

src/github.com/getlantern/flashlight/config/config.go
aranhoide: `GetCountry` returns the empty string in that case (`nil` is not a valid value for return value `string`).

https://github.com/getlantern/flashlight-build/blob/valencia/src/github.com/getlantern/flashlight/globals/globals.go#L34

src/github.com/getlantern/flashlight/config/config.go
aranhoide: It does return an error for anything but 200.

https://github.com/getlantern/flashlight-build/blob/country-specific-configs/src/github.com/getlantern/flashlight/config/config.go#L398

src/github.com/getlantern/flashlight/config/config.go
aranhoide, line 217: That's orthogonal to what is being changed here, but sure, will do.

src/github.com/getlantern/flashlight/config/config.go
aranhoide, line 391: This is meant for the client side.  I don't imagine we'll ever upload configs targetting the countries where our servers are,  Also, the string you provide is the default.  Other than that, you got it right,

If you provide a `-cloudconfig` that doesn't have a `${COUNTRY}` in it, it will just never be costomized for your particular country.

aranhoide, line 391: @atavism ^

src/github.com/getlantern/flashlight/config/config.go
aranhoide, line 217: @myleshorton : actually, changing this name will change the string in the lantern.yaml.  If we ever want to change the `cloudconfig` for old clients again we'll need to remember we need to set `cloudconfig` **and** `cloudconfigurl` to the new value so we're sure to hit old clients.  Not a huge deal, but I don't think the name is either.  I'd rather punt on this kind of stuff until we have working auto-updates and thus we don't need to care much about backwards compatibility anymore.  But if you still want it, it's a quick change.

src/github.com/getlantern/flashlight/config/config.go
myleshorton: OK cool -- thanks!

src/github.com/getlantern/flashlight/config/config.go
myleshorton, line 217: Yeah that's a good point about it also changing in the yaml itself, which would make updating old clients annoying. OK let's leave it!

src/github.com/getlantern/flashlight/config/config.go
myleshorton: Right, but even if GetCountry returns "", we're out of luck right? It'll pick it up on the next pass, but that doesn't seem ideal @aranhoide.

aranhoide: TBH I didn't think that getting it picked up on the next pass was much of a problem.  I'll make this wait for a non-"" country.

aranhoide:
@myleshorton : I just added a 20 second wait for location before trying to fetch cloud config.  This should prevent a client from running with a suboptimal configuration during its first minute ever (which is the only ill effect of the former situation AFAICT), but the timeout will still prevent a chicken-and-egg problem where a client can't geolocate itself because its current config is preventing domain fronting from working.

myleshorton:
Thanks @aranhoide. I mostly just think that first minute is a pretty crucial minute given that it's your very first impression of Lantern.

I think it's almost tempting to make GetLocation and GetCountry block until there's an actual value. It's hard to imagine anything calling those methods that wouldn't require having an actual value.

aranhoide:
@myleshorton : I tried and it's a can of worms.  For one thing, geolookup itself calls GetLocation in a couple places.  And so far I couldn't keep a spurious "xx" location from creeping in at startup.  I'd say if we really want to make this change let's file an issue for it and address it in its own PR.

myleshorton:
Sounds good I think @aranhoide. The other simple approach I meant to mention was to just use the observer/listener pattern where listeners just get notified whenever the location is populated. For now this would be the only listener.

myleshorton:
OK, merging. Thanks @aranhoide.

aranhoide:
Thanks for reviewing!  As for doing a pub/sub scheme, that would probably be the most proper way to implement this, but again I think that's best done in a ticket of its own, since it's more of a change than making GetLocation/GetCountry block.  BTW, I think there would be at least another listener: the system that renders your dot in the map.

PR #171, opened 2015-04-10
make cloud config completely overwrite servers and masquerade sets

aranhoide:
It used to add any new servers and masquerades, keeping the old ones in place.
This means we couldn't remove dead or slow servers nor masquerades from the
cloud config.

Fixes https://github.com/getlantern/lantern/issues/2365

### DIFF
--- a/src/github.com/getlantern/flashlight/config/config.go
+++ b/src/github.com/getlantern/flashlight/config/config.go
@@ -388,26 +388,25 @@ func (cfg Config) doFetchCloudConfig(proxyAddr string) ([]byte, error) {
 	return ioutil.ReadAll(gzReader)
 }
 
-// updateFrom creates a new Config by merging the given yaml into this Config.
-// Any servers in the updated yaml replace ones in the original Config and any
-// masquerade sets in the updated yaml replace ones in the original Config.
+// updateFrom creates a new Config by 'merging' the given yaml into this Config.
+// The masquerade sets and the collections of servers in the update yaml
+// completely replace the ones in the original Config.
 func (updated *Config) updateFrom(updateBytes []byte) error {
+	// XXX: does this need a mutex, along with everyone that uses the config?
+	oldFrontedServers := updated.Client.FrontedServers
+	oldChainedServers := updated.Client.ChainedServers
+	oldMasqueradeSets := updated.Client.MasqueradeSets
+	updated.Client.FrontedServers = []*client.FrontedServerInfo{}
+	updated.Client.ChainedServers = map[string]*client.ChainedServerInfo{}
+	updated.Client.MasqueradeSets = map[string][]*fronted.Masquerade{}
 	err := yaml.Unmarshal(updateBytes, updated)
 	if err != nil {
+		updated.Client.FrontedServers = oldFrontedServers
+		updated.Client.ChainedServers = oldChainedServers
+		updated.Client.MasqueradeSets = oldMasqueradeSets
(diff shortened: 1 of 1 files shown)


src/github.com/getlantern/flashlight/config/config.go
```diff
@@ -388,26 +388,25 @@ func (cfg Config) doFetchCloudConfig(proxyAddr string) ([]byte, error) {
...
 }
 
-// updateFrom creates a new Config by merging the given yaml into this Config.
-// Any servers in the updated yaml replace ones in the original Config and any
-// masquerade sets in the updated yaml replace ones in the original Config.
+// updateFrom creates a new Config by 'merging' the given yaml into this Config.
+// The masquerade sets and the collections of servers in the update yaml
+// completely replace the ones in the original Config.
 func (updated *Config) updateFrom(updateBytes []byte) error {
+	// XXX: does this need a mutex, along with everyone that uses the config?
```
myleshorton, line 395: I believe it's only ever called serially with the updated copy getting passed to all code that uses it, but it's a really good question since clearly things like balancer are using these directly. It's honestly a bit hard to trace all the code paths, but I'll keep looking at it.

We decided not to cut beta5, by the way, as none of us was quite able to get to these code reviews. Thanks for getting them in though. Hopefully it'll go out Monday.

src/github.com/getlantern/flashlight/config/config.go
```diff
@@ -388,26 +388,25 @@ func (cfg Config) doFetchCloudConfig(proxyAddr string) ([]byte, error) {
 	return ioutil.ReadAll(gzReader)
 }
 
-// updateFrom creates a new Config by merging the given yaml into this Config.
-// Any servers in the updated yaml replace ones in the original Config and any
-// masquerade sets in the updated yaml replace ones in the original Config.
+// updateFrom creates a new Config by 'merging' the given yaml into this Config.
+// The masquerade sets and the collections of servers in the update yaml
+// completely replace the ones in the original Config.
 func (updated *Config) updateFrom(updateBytes []byte) error {
```
myleshorton, line 394: I do think a test for this method would be extremely valuable.

src/github.com/getlantern/flashlight/config/config.go
myleshorton: I gotta say the option @fffw raised of just making the cloud config and user configs separate is appealing to me @aranhoide. We could also do the following that would make testing much simpler:

1. Always have the cloud.yaml reflect the global cloud.yaml
1. Have a separate local-cloud.yaml file that would purely be for testing/development that would completely overwrite cloud.yaml
1. In addition have a separate local-config.yaml file that would exclusively have settings that are only relevant for the client side.

I feel like that makes things significantly simpler conceptually. The names of all of the above files are not at all fixed of course.

What do you think? Could also happen in a future pass, but given that we have almost no user settings now it might be an easy time to change it too.

myleshorton, line 395: I confirmed we don't need a lock @aranhoide, as each config file that's propagated is a copy and is effectively read only. This particular chunk of code is pretty much the first thing to touch it too.

myleshorton, line 395: Just added a ticket for the above:

https://github.com/getlantern/lantern/issues/2390

myleshorton:
OK, LGTM for the purposes of this pass.

aranhoide:
Thanks!

PR #172, opened 2015-04-13
Flashlight & Lantern mobile updates

atavism:
This includes changes for https://github.com/getlantern/lantern/issues/2363 to consolidate flashlight and the lantern android library further.

With the latest changes, I restored configuration polling and fixed some issues with the unit tests.


xiam:
Sounds awesome Todd! I'll take a look into it.

src/github.com/getlantern/lantern-android/client/client.go
```diff
@@ -1,29 +1,32 @@
...
 	cloudConfigPollInterval = time.Second * 60
-	httpConnectMethod       = "CONNECT"
-	httpXFlashlightQOS      = "X-Flashlight-QOS"
 )
 
 // clientConfig holds global configuration settings for all clients.
 var clientConfig *config
 
+// MClient is an extension of flashlight client with a few custom declarations for mobile
+type MClient struct {
```
xiam: It is a bit unusual to abbreviate words in Go code when not using acronyms such as `HTTP`, at least in the standard library. I think renaming `MClient` to `MobileClient` or just `Mobile` would improve readability. What do you think?

src/github.com/getlantern/lantern-android/client/client.go
```diff
@@ -124,141 +120,17 @@ func (client *Client) pollConfiguration() {
...
-
-func respondOK(writer io.Writer, req *http.Request) error {
-	defer req.Body.Close()
-	resp := &http.Response{
-		StatusCode: http.StatusOK,
-		ProtoMajor: 1,
-		ProtoMinor: 1,
-	}
-	return resp.Write(writer)
+	return client.hqfd.Close()
```
xiam: Does this line stop the client that was spawned in `Client.ServeHTTP`?

src/github.com/getlantern/lantern-android/client/client_test.go
```diff
@@ -27,14 +27,7 @@ func TestListenAndServeStop(t *testing.T) {
...
-	}
-
-	// Make the client listen on a goroutine.
-	go func() {
-		c.ListenAndServe()
-	}()
+	c.ServeHTTP()
 
 	// Allow it some seconds to start.
 	time.Sleep(time.Millisecond * 100)
```
xiam, line 33: Didn't realize this before but we can use waitforserver here.

xiam, line 33: Forget it, it's a test :-)

xiam:
The `src/github.com/getlantern/lantern-android/app/libs/armeabi-v7a/libgojni.so` library is a binary file that gets rebuilt every time, I guess we should not include it in the source three. Maybe we should remove it add it to the `.gitignore`?

src/github.com/getlantern/lantern-android/client/client.go
```diff
@@ -1,29 +1,32 @@
...
-	httpXFlashlightQOS      = "X-Flashlight-QOS"
 )
 
 // clientConfig holds global configuration settings for all clients.
 var clientConfig *config
 
+// MClient is an extension of flashlight client with a few custom declarations for mobile
+type MClient struct {
+	client.Client
+	hqfd   fronted.Dialer
```
xiam: What's the main purpose of `hqfd` being a member of `MClient`?

src/github.com/getlantern/lantern-android/client/client_test.go
```diff
@@ -50,10 +43,13 @@ func TestListenAndServeAgain(t *testing.T) {
 	// Since we've closed out server, we should be able to launch another at the
 	// same address.
 
+	globalClient = NewClient(listenProxyAddr)
 	go func() {
 		globalClient = NewClient(listenProxyAddr)
 
-		if err := globalClient.ListenAndServe(); err != nil {
+		onListening := func() {}
```
xiam: What is the purpose of the empty function?

src/github.com/getlantern/lantern-android/client/client.go
atavism: To be able to run `StopClientProxy()` with the sample program since flashlight client doesn't explicitly expose a stop method. `hqfd.Close()`, which is called in `Stop()`, closes the underlying connection pool.

xiam:
Code looks good @atavism! Thanks, I added only a few comments, now I need to verify if everything is still working in both the emulator and a real device.

src/github.com/getlantern/lantern-android/client/client.go
atavism: I agree. I just renamed it to MobileClient.

src/github.com/getlantern/lantern-android/client/client.go
atavism: Yes, it's intended to. I'm basically just supporting the Stop functionality you provided previously.

src/github.com/getlantern/flashlight/client/client.go
```diff
@@ -99,3 +101,10 @@ func (client *Client) Configure(cfg *ClientConfig) fronted.Dialer {
 
 	return client.hqfd
 }
+
+// Stop is called when the client is no longer needed. It closes the
+// client listener and underlying dialer connection pool
+func (client *Client) Stop() error {
+	client.hqfd.Close()
+	return client.l.Close()
+}
```
atavism, line 110: @xiam To make it cleaner, I added `Stop()` to flashlight client.

src/github.com/getlantern/lantern-android/client/client.go
atavism: I moved `Stop()` to flashlight client and removed `hqfd` from `MobileClient`

xiam:
@atavism I tried to test on a real device, but it seems like the library pulled some unwanted stuff, in particular the `os/user` package that is [not implemented](http://golang.org/src/os/user/lookup_stubs.go) for ARM devices:

```
04-13 14:03:28.525  12090-12116/org.getlantern.example E/Go﹕ panic: Unable to construct byteexec for upnpc: Unable to determine user's home directory: user: Current not implemented on android/arm
04-13 14:03:28.525  12090-12116/org.getlantern.example E/Go﹕ [ 04-13 14:03:28.525 12090:12116 E/Go       ]
    goroutine 1 [running]:
04-13 14:03:28.526  12090-12116/org.getlantern.example E/Go﹕ github.com/getlantern/go-igdman/igdman.init·1()
04-13 14:03:28.526  12090-12116/org.getlantern.example E/Go﹕ /src/github.com/getlantern/go-igdman/igdman/upnp.go:34 +0x294
04-13 14:03:28.526  12090-12116/org.getlantern.example E/Go﹕ github.com/getlantern/go-igdman/igdman.init()
```

I remember that was the reason for cloning the Config type.

This must be debugged on a Linux host, I can't test right now but this probably won't work on Linux:

```
cd src/github.com/getlantern/lantern-android/client
go test -tags=android
```

Did you deploy successfully to a real device? Of what type?

atavism:
I'm not seeing that at all. Where are you putting lingojni.so, and are you using the latest golang/mobile? Yeah, it's been working fine for me in the emulator and on a real device.. for @ffww too. I've been testing arm and armv7. What about the shared lib I already committed? Does that work for you?

atavism:
Also, when I grep the *.so, I don't get any matches for "upnp"--are we maybe building the library differently somehow?

xiam:
@atavism 

This is really odd. I'm using this one:

```
shasum -a 256 libgojni.so
427a61009194c52d11f813aed00a0edf50281c8f7d1941093dabad28fca42cba  libgojni.so
strings libgojni.so | grep upnp | wc -l
      48
```

I think I'm up to date:

```
git log | head -n1
commit 0f8f24ed0fb1ed9c6ffa3ba57181a83fd782ad32
```

I'm deploying an Ubuntu VM to test if running the test with the arm tag works, I think it will fail but let's find why this is not working.

Do you have the same revision? Are you sure you're using the same file in your Android project's `app/src/main/jniLibs/armeabi-v7a` directory?

xiam:
This looks better from within a Linux box, we have to cheat the `go build` command a bit in order for it to build with the `android` tag, just like a normal build for android but with the current OS/Arch. It fails too:

```
CGO_ENABLED=0 go test -tags android
flashlight.ui: Creating tarfs filesystem that prefers local resources at /vagrant/flashlight-build/src/github.com/getlantern/lantern-ui/app
tarfs: Using local filesystem at /vagrant/flashlight-build/src/github.com/getlantern/lantern-ui/app
panic: Unable to construct byteexec for upnpc: Unable to determine user's home directory: user: Current not implemented on linux/amd64

goroutine 1 [running]:
github.com/getlantern/go-igdman/igdman.init·1()
        /vagrant/flashlight-build/src/github.com/getlantern/go-igdman/igdman/upnp.go:34 +0x2f7
github.com/getlantern/go-igdman/igdman.init()
        /vagrant/flashlight-build/src/github.com/getlantern/go-igdman/igdman/util_linux.go:35 +0x164
github.com/getlantern/flashlight/server.init()
        /vagrant/flashlight-build/src/github.com/getlantern/flashlight/server/server.go:436 +0x68
github.com/getlantern/flashlight/config.init()
        /vagrant/flashlight-build/src/github.com/getlantern/flashlight/config/proxiedsites.go:20 +0x63
github.com/getlantern/flashlight/analytics.init()
        /vagrant/flashlight-build/src/github.com/getlantern/flashlight/analytics/analytics.go:125 +0x5e
github.com/getlantern/lantern-android/client.init()
        /vagrant/flashlight-build/src/github.com/getlantern/lantern-android/client/config_test.go:40 +0x9f
main.init()
        github.com/getlantern/lantern-android/client/_test/_testmain.go:64 +0x4a
exit status 2
FAIL    github.com/getlantern/lantern-android/client    0.029s
```

xiam:
I also tested the last .so:

```
strings libgojni.so | grep upnp | wc -l
       0
shasum -a 256 libgojni.so
57d5f4f515540061e0e345c0ee46122bd0cc3278ede02343b0d6e19f0727d050  libgojni.so
```

It also issues a panic like this:

```
04-13 14:03:28.507  12090-12119/org.getlantern.example I/GoStdio﹕ flashlight.ui: Creating tarfs filesystem that prefers local resources at /src/github.com/getlantern/lantern-ui/app
04-13 14:03:28.525  12090-12116/org.getlantern.example E/Go﹕ panic: Unable to construct byteexec for upnpc: Unable to determine user's home directory: user: Current not implemented on android/arm
04-13 14:03:28.525  12090-12116/org.getlantern.example E/Go﹕ [ 04-13 14:03:28.525 12090:12116 E/Go       ]
    goroutine 1 [running]:
04-13 14:03:28.526  12090-12116/org.getlantern.example E/Go﹕ github.com/getlantern/go-igdman/igdman.init·1()
04-13 14:03:28.526  12090-12116/org.getlantern.example E/Go﹕ /src/github.com/getlantern/go-igdman/igdman/upnp.go:34 +0x294
04-13 14:03:28.526  12090-12116/org.getlantern.example E/Go﹕ github.com/getlantern/go-igdman/igdman.init()
```

But unlike the other version of the `.so`, this one seems to continue working after the panic, which is really interesting...

atavism:
@xiam Yeah, I'm trying to figure out what the exact problem is. It's quite strange: the byteexec/upnp stuff doesn't always get pulled into the build. When it's totally absent, as you pointed out, it continues working fine (and that's the version FireTweet is using right now).

xiam:
I also tested building with the Makefile but it builds a library with upnp.

The problem here is not only upnp but the call to `appdir.General` within flashlight's `config/config.go`, `appdir` is currently not compatible with Go Android, because of the call to `user.Current()` (which does not make sense on Android OS).

atavism:
@xiam Hmm, not sure where anything in `config/config.go` is actually called with my updates. That said, what do you think about just modifying the `panic: Unable to construct byteexec for upnpc: Unable to determine user's home directory: user: Current not implemented on linux/amd64` line in `upnp.go` to log an error instead of panicking? Or maybe disable that check if the architecture is ARM? We aren't even calling any of that code anywhere so it shouldn't affect anything.

xiam:
Hello @atavism,

The `InConfigDir` function called inside `config/config.go` calls to `appdir.General` which depends on `os/user`. We're doing the same thing on logging `logging/logging.go`, so we can have undefined behaviour on those two packages. 

This is the list of Lantern packages that depend on 'os/user':

```
grep -R 'os/user' .
./appdir/appdir.go: "os/user"
./aws-sdk-go/aws/auth.go:   "os/user"
./byteexec/byteexec.go: "os/user"
./keyman/keyman_trust_linux.go: "os/user"
```

The source of the panic seems to be `./go-igdman/igdman/upnp.go` on line 34:

```
panic(fmt.Errorf("Unable to construct byteexec for upnpc: %s", err))
```

Now, it looks like you're not importing `igdman` directly but as a dependency of other packages:

The `github.com/getlantern/flashlight/analytics` package imports `github.com/getlantern/flashlight/config` which imports `github.com/getlantern/flashlight/server` which imports `github.com/getlantern/go-igdman/igdman`. 

The `github.com/getlantern/flashlight/analytics` package also imports `github.com/getlantern/flashlight/config` which imports `github.com/getlantern/appdir`.

Finally, `github.com/getlantern/flashlight/globals` imports `github.com/getlantern/keyman` which imports `os/user`. Also note that  `getlantern/flashlight/client` imports both `github.com/getlantern/flashlight/globals` and `github.com/getlantern/keyman`.

The problem is only getting bigger, we could start adding runtime conditionals and build tags but that will surely provide some inestability to our development branch, for we'll have to touch a lot of code, that's the reason I decided to just clone Config and implement a Client.

What to do? I think it would be safer to remove `getlantern/flashlight/analytics`, `getlantern/flashlight/globals` and `getlantern/flashlight/client` from the code and try to make Lantern-Android work with some of the most essential pieces it needs instead of starting changing our codebase to make Lantern-Android happy. The proxy part of Flashlight is not that big, so we'll still have some duplicated code but nothing too large to maintain at this time.

Talking about the technical debt @myleshorton pointed out yesterday, we might at some point go back and fix these issues:
- `config` should not be importing `igdman`.
- `config` should not be importing `server`.
- `config` should not depend on any `client` or `server` code.
- Lantern's `server` and `client` libraries must be independent code entities.
- Lantern must be prepared to handle Android-like scenarios in which we don't have a $HOME or even a user name.
- Lantern must be prepared to not being able to read or write to disk.

And then doing this lantern-android consolidation would be easier.

What do you think @atavism, @myleshorton?

atavism:
Hey @xiam,

Thanks a lot for listing all of the dependencies and import paths that are affecting the latest changes. 

We aren't using `logging/logging.go` anywhere in Lantern-Android so it seems to me the main issue here as you pointed out is with `go-igdman/igdman/upnp.go` which panics when byteexec tries to determine the user home directory and is unable to write upnp anywhere. I agree we shouldn't add special checks to accommodate Lantern-Android, but I feel we should try to resolve this now instead of reverting to what we already had. I actually think there's an even easier solution: stop importing `config` in the analytics package. There's no reason we have to (I only did so as a convenience), and that would prevent inclusion of `go-igdman` entirely. That's actually a small and useful change overall anyway. After that, I agree we should move forward and fix the issues you summarized. How does that sound?

xiam:
@atavism Sounds like a good alternative to me. The `client` package will still have references to `github.com/getlantern/keyman` which might be affected because of `os/user` but we can try and see if that does the trick.

xiam:
Hello @atavism,

I saw a message in the analytics implementation:

```
        // for now, the only analytics messages we are
        // currently receiving from the UI are initial page
        // views which indicate new UI sessions
```

What's the current case for Android and Analytics?

atavism:
@xiam We track sessions now too. I went ahead and removed the `analytics.Configure()` call and the `flashlight/analytics` import. Do you want to try testing the library again with the latest changes?

xiam:
@atavism Awesome! the last build you uploaded is working and it does not show any panic at all, the local build I created too, but for some reason on both cases the proxy takes some time to start listening... Still haven't discovered why.

src/github.com/getlantern/lantern-android/app/src/go/Seq.java
```diff
@@ -36,6 +36,7 @@ public Seq() {
 	public native float readFloat32();
```
xiam: This file is generated by the build process, we should probably not take it into the source tree as it is not really editable.

src/github.com/getlantern/lantern-android/app/src/org/getlantern/Flashlight.java
```diff
@@ -12,9 +12,9 @@ private Flashlight() {} // uninstantiable
     public static void RunClientProxy(String listenAddr) throws Exception {
```
xiam: This file is generated by the build process, we should probably not take it into the source tree as it is not really editable.

src/github.com/getlantern/lantern-android/libflashlight/bindings/go_bindings/go_bindings.go
```diff
@@ -10,21 +10,21 @@ import (
 )
```
xiam: This file is generated by the build process, we should probably not take it into the source tree as it is not really editable.

xiam:
Overall this looks good @atavism, I also tested the library that I build locally. Looks fine. I'm only worried about the files that are auto generated, including the library `libgojni.so`, I think they add some noise as they are not editable and can be easily build from the source code by using a script.

atavism:
@xiam Thanks for reviewing! I totally agree we don't need to commit the shared object or files that were auto-generated. I mainly added them for convenience because I frequently needed to copy them over elsewhere. I went ahead and removed them from the PR.

xiam:
Great stuff @atavism, thanks!

PR #174, opened 2015-04-13
Using a lower-cased lantern config directory on Linux.

xiam:
I realized the "Lantern" name with the first character in uppercase looks fine on OSX: `/Users/rev/Library/Application\ Support/Lantern` and other apps that are in `/Users/rev/Library/Application\ Support/` follow the same convention. I just added a case for Linux, in which having cases looks and feels odd.

### DIFF
--- a/src/github.com/getlantern/flashlight/config/config.go
+++ b/src/github.com/getlantern/flashlight/config/config.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
+	"runtime"
 	"sort"
 	"time"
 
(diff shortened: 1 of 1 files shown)


src/github.com/getlantern/flashlight/config/config.go
```diff
@@ -155,18 +156,39 @@ func Update(mutate func(cfg *Config) error) error {
...
 	if cdir == "" {
-		cdir = appdir.General("Lantern")
+		if runtime.GOOS == "linux" {
+			// It is more common on Linux to expect application related directories
+			// in all lowercase. The lantern wrapper also expects a lowercased
+			// directory.
+			cdir = appdir.General("lantern")
+
+			// Backwards compatibility fix.
+			oldDir := appdir.General("Lantern")
```
myleshorton: I actually don't think we should worry about backwards compatibility for this at this point. This has still only been released in beta, and there are almost no user-configurable settings. So I would skip adding this complexity at this point.

src/github.com/getlantern/flashlight/config/config.go
```diff
@@ -155,18 +156,39 @@ func Update(mutate func(cfg *Config) error) error {
...
+			cdir = appdir.General("Lantern")
+		}
 	}
+
 	log.Debugf("Placing configuration in %v", cdir)
 	if _, err := os.Stat(cdir); err != nil {
 		if os.IsNotExist(err) {
 			// Create config dir
-			if err := os.MkdirAll(cdir, 0755); err != nil {
+			if err := os.MkdirAll(cdir, 0750); err != nil {
```
xiam, line 176: I changed this to 750 to prevent other users or applications from other users from reading current and future Lantern resources.

src/github.com/getlantern/flashlight/config/config.go
xiam: @myleshorton Agreed! I removed the unnecessary part.

myleshorton:
LGTM!

PR #175, opened 2015-04-13
Add setting to run Lantern on system startup

atavism:
https://github.com/getlantern/lantern/issues/2344


myleshorton:
It looks like this still has all the windows service code from code.google.com/p/winsvc in there even though it's not used @atavism .

atavism:
@myleshorton Ah, thanks. I just removed it.

src/github.com/getlantern/launcher/launcher_windows.go
```diff
@@ -0,0 +1,38 @@
...
+func CreateLaunchFile(autoLaunch bool) {
+	var err error
+
+	if autoLaunch {
+		lanternPath, err := osext.Executable()
+		if err != nil {
+			log.Errorf("Could not get Lantern directory path: %q", err)
+			return
+		}
+		err = gowin.WriteStringReg("HKCU", RunDir, "value", lanternPath)
```
myleshorton: These registry calls still succeed if not running as admin, is that right? Might be worth testing with a full installer.

atavism: Yeah, I verified with regedit. Good point though we should test these changes with the full installer though

src/github.com/getlantern/launcher/launcher_darwin.go
```diff
@@ -0,0 +1,59 @@
...
+	t := template.Must(template.New("LaunchdPlist").Parse(LaunchdPlist))
+
+	err = t.Execute(&content, &Plist{RunAtLoad: autoLaunch})
+	if err != nil {
+		log.Errorf("Error writing plist template: %q", err)
+		return
+	}
+
+	if err = ioutil.WriteFile(fname, content.Bytes(), 0755); err != nil {
+		log.Fatalf("Error writing to launchd plist file: %q", err)
```
myleshorton: I don't think we want this to be fatal here @atavism. The old lantern unfortunately would create this file with root permissions, so for me writing to it failed just now, and Lantern completely died. So I think we should should just log the error.

To avoid this conflict we could also name the file something slightly different like org.getlantern.plist

myleshorton:
I get this build error on windows:

```
src/github.com/getlantern/launcher/launcher_windows.go:6:2: no buildable Go source files in /flashlight-build/src/github.com/luisiturrios/gowin
```

Indeed that package is empty!

src/github.com/getlantern/launcher/launcher_darwin.go
atavism: Ah, I didn't realize the old Lantern wrote the plist file with root permissions. That sounds good--I just made the changes you suggested.

myleshorton:
Shoot now I get:

```
src/github.com/luisiturrios/gowin/registry.go:8:2: cannot find package "code.google.com/p/winsvc/winapi" in any of:
```

So sorry looks like gowin actually uses winsvc? Seems a little unfortunate but probably OK.

src/github.com/getlantern/launcher/launcher_windows.go
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+	"github.com/kardianos/osext"
+	"github.com/luisiturrios/gowin"
+
+	"github.com/getlantern/golog"
+)
+
+const (
+	RunDir = `Software\Microsoft\Windows\CurrentVersion\Run`
```
myleshorton: nit: it seems like this should just be a local variable in CreateLaunchFile since it's not used anywhere else and just adds to the global state otherwise.

atavism: Yeah, true. I usually prefer declaring constants at the top of a file regardless, but I went ahead and moved it inside the function body.

teamlantern: Cool yeah I guess its partly stylistic.

myleshorton:
Thanks on the winsvc front. I still am not sure what's going on, but if I build the installer from this branch using:

```
VERSION=2.0.0-beta5 make package-windows
```

The resulting installer does two odd things:
1. It doesn't have a settings button
2. It goes directly to the old welcome screen

Both make it seem like it's using an old version of the UI. Do you see the same things @atavism? This could also be related to cache settings -- it's possible the browser's just caching the old copy of the UI.

Oh, also, I can't see to get it to affect the registry. I also checked, and there aren't any stray 2.0.0-beta5 tags lying around that might make it point to the wrong version.

atavism:
@myleshorton Yeah, this was happening to me too at some point. Did tarfs update the `resources.go` file after you built the installers? Also, you might try running `UPDATE_DIST=true make docker-genassets` prior to that (though the latest changes should already be committed and I see them in the diff).

myleshorton:
OK @atavism this issue with the outdated UI on Windows is gone for me now, but it's unfortunately still not successfully updating the registry when I toggle the auto-start setting.

myleshorton:
Actually it looks like the auto-start checkbox is controlling stats reporting for me. The logs show:

```
flashlight.statreporter: Stat reporting turned off
```

Oddly clicking either checkbox on or off seems to always turn **stat reporting off**. This is also in Firefox so could be related to that and not the browser. I'll check Chrome on Windows.

atavism:
@myleshorton You pulled in the latest changes, right? Are you looking for the registry key in `Software\Microsoft\Windows\CurrentVersion\Run` under HKCU?

atavism:
It's not actually turning stat reporting on/off.. that just means config updates are propagating. those messages are always printed.

myleshorton:
Hey thanks @atavism. Actually, it looks like it's correctly turning it _on_ but just not _off_.

atavism:
@myleshorton Thanks. Just fixed another issue with the path we were using.
- Here are the two tests I'm running in addition to checking the registry key gets added/removed. Note: the run on system start setting is disabled by default

```
- check "Run Lantern on system start" in Settings 
  Quit Lantern 
  Logout and log back in 
  -> Lantern should start
```

```
- uncheck "Run Lantern on system start" in Settings
  Quit Lantern
  Logout and log back in 
  -> Lantern shouldn't start at all
```

myleshorton:
OK this is working for me. I'm going to merge since the test failure has been fixed on valencia (ui resources again).

PR #178, opened 2015-04-15
Fixing compilation problem with default resources file. (Issue #2388)

xiam:
The problem was we weren't able to build Lantern without make genassets. This should close https://github.com/getlantern/lantern/issues/2388.


myleshorton:
Nice @xiam!! Looks great. Merging.

PR #179, opened 2015-04-15
Patch for PR#175

xiam:
Following https://github.com/getlantern/flashlight-build/pull/175

### DIFF
--- a/src/github.com/getlantern/flashlight/flashlight.go
+++ b/src/github.com/getlantern/flashlight/flashlight.go
@@ -110,7 +110,7 @@ func doMain() error {
 	displayVersion()
 
 	parseFlags()
-	configUpdates = make(chan *config.Config)
+
 	cfg, err := config.Init()
 	if err != nil {
 		return fmt.Errorf("Unable to initialize configuration: %v", err)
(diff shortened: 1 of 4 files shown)


src/github.com/getlantern/flashlight/flashlight.go
```diff
@@ -110,7 +110,7 @@ func doMain() error {
 	displayVersion()
 
 	parseFlags()
-	configUpdates = make(chan *config.Config)
```
xiam, line 113: We didn't really need to realloc configUpdates.

src/github.com/getlantern/launcher/launcher_windows.go
```diff
@@ -8,13 +8,16 @@ import (
 	"github.com/getlantern/golog"
 )
 
+const (
+	runDir = `Software\Microsoft\Windows\CurrentVersion\Run`
```
xiam, line 12: Instead of converting this constant into a variable scoped to a single function I made it an unexported constant by making the first char lowercase. It would be easier to become aware of its _constantness_ this way.

atavism, line 12: Yeah, that's actually identical to what I had originally =P @myleshorton recommended putting it inside the only function it's used in. I don't think it makes a big difference, but I prefer making it clear its a constant too.

atavism:
This is working great! Merging

myleshorton:
Nice! LGTM too.

PR #185, opened 2015-04-21
Route HTTP traffic to trusted proxies only

xiam:
This PR adds support for routing plaintext HTTP traffic only to proxies deployed by Lantern, in this moment it may not represent any difference.

HTTPs can travel safely along trusted and untrusted proxies without compromising user's data, as the protocol itself provides point to point encryption (well, point to Trusted Third Parties and eventually to point, but that's outside our scope).

A new property `trusted` was added to our configuration settings, this feature depends on whether this flag is set true or not. If no servers marked as `trusted` are found Lantern will try to use any server (getting unsafe access is better than having no access at all).

Routing an HTTP URL would produce a log like:

```
curl -x 127.0.0.1:8787 http://www.google.com/humans.txt

Apr 21 19:56:19.621 - balancer: balancer.go:100 Dialing tcp://www.google.com:80 with (trusted) fronted proxy at nl.fallbacks.getiantem.org:443 using masquerade set cloudflare
```

While routing an HTTPs URL would look like:

```
curl -x 127.0.0.1:8787 https://www.google.com/humans.txt

Apr 21 19:57:05.401 - balancer: balancer.go:100 Dialing tcp://s3.amazonaws.com:443 with chained proxy at 178.62.243.211:443
```

You'll probably need to edit `lantern.yaml` manually to set the `trusted` field on fronted and chained proxies.

I also added many comments on the code that would make it easier to understand what is each part trying to do and fixed a few minor bugs.

See our discussion at: https://github.com/getlantern/lantern/issues/2290


src/github.com/getlantern/balancer/balancer.go
```diff
@@ -22,23 +22,42 @@ var (
 // Balancer balances connections established by one or more Dialers.
 type Balancer struct {
 	dialers []*dialer
+	trusted []*dialer
```
xiam: This property will be nil until calling `trustedDialers`.

src/github.com/getlantern/balancer/balancer.go
```diff
@@ -49,7 +68,25 @@ func New(dialers ...*Dialer) *Balancer {
 // remaining Dialers until it either manages to connect, or runs out of dialers
 // in which case it returns an error.
 func (b *Balancer) DialQOS(network, addr string, targetQOS int) (net.Conn, error) {
```
xiam, line 70: This is the only function that returns a `net.Conn`, so it looks like a good place to check if the request is unencrypted and take action.

src/github.com/getlantern/chained/server.go
```diff
@@ -69,7 +69,7 @@ func (s *Server) ServeHTTP(resp http.ResponseWriter, req *http.Request) {
 		wg.Done()
 	}()
 	go func() {
-		go io.Copy(connIn, connOut)
+		io.Copy(connIn, connOut)
```
xiam, line 72: The `wg.Wait()` was not actually doing nothing as the `io.Copy` was spawned into a goroutine. Now the `wg.Wait()` will actually wait that both calls to `io.Copy()` are completed before returning control to the parent function.

src/github.com/getlantern/flashlight/client/balancer.go
```diff
@@ -7,40 +7,70 @@ import (
...
 	for _, s := range cfg.FrontedServers {
+		// Getting a dialer for domain fronting and a dialer to dial to arbitrary
+		// addreses.
 		fd, dialer := s.dialer(cfg.MasqueradeSets)
 		dialers = append(dialers, dialer)
 		if dialer.QOS > highestQOS {
+			// If this dialer as a higher QOS than our current highestQOS, set it as
+			// the highestQOSFrontedDialer.
 			highestQOSFrontedDialer = fd
+			highestQOS = dialer.QOS
```
xiam, line 49: We were missing a line that updated `highestQOS` to the greatest known value of `dialer.QOS` in order to actually get the highest QOS.

src/github.com/getlantern/flashlight/client/balancer.go
```diff
@@ -7,40 +7,70 @@ import (
...
 			log.Debugf("Unable to configure chained server for %s: %s", s.Addr)
 		}
-
 	}
 
+	// Creating a balancer with all of our available dialers.
 	bal := balancer.New(dialers...)
 
+	// Locking balCh.
+	client.balChMu.Lock()
```
xiam, line 69: There's a small change that both `initBalancer` and `getBalancer` would push something to `balCh`. Since `balCh` is a channel of just one element it would block. It would be a very rare condition to reproduce.

src/github.com/getlantern/flashlight/client/handler.go
```diff
@@ -32,39 +34,59 @@ func (client *Client) ServeHTTP(resp http.ResponseWriter, req *http.Request) {
...
+	if req.Method != httpConnectMethod {
 		panic("Intercept used for non-CONNECT request!")
 	}
 
-	// Hijack underlying connection
-	clientConn, _, err := resp.(http.Hijacker).Hijack()
-	if err != nil {
+	// Hijacking underlying connection.
+	var clientConn net.Conn
+	if clientConn, _, err = resp.(http.Hijacker).Hijack(); err != nil {
```
xiam, line 46: Using this error declaration pattern and comparison is great for quickly identifying which error are we actually referring to.

src/github.com/getlantern/flashlight/client/handler.go
```diff
@@ -11,18 +11,20 @@ import (
 )
 
 const (
-	CONNECT        = "CONNECT" // HTTP CONNECT method
-	XFlashlightQOS = "X-Flashlight-QOS"
```
xiam, line 15: No need to export those constants.

src/github.com/getlantern/flashlight/client/handler.go
```diff
@@ -78,51 +100,41 @@ func (client *Client) targetQOS(req *http.Request) int {
...
 		ProtoMinor: 1,
 	}
-	err := resp.Write(w)
+	err = resp.Write(w)
 	if err == nil {
 		_, err = w.Write([]byte(msg))
 	}
 	return err
 }
-
```
xiam, line 118: We don't really need this as a separate function. The intention of this code is more evident if we move it to the context in which it's actually being used.

src/github.com/getlantern/flashlight/client/reverseproxy.go
```diff
@@ -38,22 +51,38 @@ func (client *Client) initReverseProxy(bal *balancer.Balancer, dumpHeaders bool)
...
 	log.Trace("Publishing reverse proxy")
+
+	// getReverseProxy() will be unblocked after this.
 	client.rpCh <- rp
+
+	// Unlocking rpCh.
+	client.rpChMu.Unlock()
+
+	// Setting the rpInitialized flag.
+	client.rpInitialized = true
```
xiam, line 85: We were not actually recording if this was initialized.

